PR #27, opened 2014-11-24
Returning found builds

HNygard:
Looked at the code for getting downstream builds and found this method that did not return anything, ever. Changed the returned statement to the way I think it should be. Have not tested the code.

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/dbf/CoreDBF.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/dbf/CoreDBF.java
@@ -70,6 +70,6 @@ public class CoreDBF extends DownstreamBuildFinder {
             }
         }
 
-        return EMPTY;
+        return foundBuilds;
     }
 }


AnneTheAgile:
@HNygard , did you make this change on your own version and it's working/helping you ? The doc ref'd by that auto-comment says you should escalate to the dev list if you don't hear. There may not be anyone listening in for PR's. http://jenkins-ci.org/content/mailing-lists

HNygard:
@AnneTheAgile : No, I have not tested this commit. And I have not sent any emails. Somebody can merge it if the find it :-)

rsandell:
It looks like it should be correct.
I was waiting and hoping for @JOSivtoft would take a look since he had a better comprehension of the context of this. Or maybe @TWestling could catch him at the water cooler and ask.

JOSivtoft:
I have looked into the code, but honestly i don't remember any reason for having it return EMPTY. The change looks reasonable. I will talk to @TWestling next week when he returns to the office. I don't have any environment set up for Jenkins nowadays.

TWestling:
I agree, this really looks like a typo, the foundBuilds are calculated but never returned.

PR #29, opened 2014-12-19
[FIXED JENKINS-24434] fix trim() usage

ilkerunaldi:
bug 24434 is caused by removal of all whitespace by trim() function. it looks like only eol character removal was intended.

fix tested & verified with jenkins 1.528.

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/IndicationAnnotator.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/IndicationAnnotator.java
@@ -77,7 +77,7 @@ public class IndicationAnnotator extends ConsoleAnnotator<Object> {
 
     @Override
     public ConsoleAnnotator annotate(Object context, MarkupText text) {
-        AnnotationHelper match = helperMap.get(text.getText().trim());
+        AnnotationHelper match = helperMap.get(text.getText().replace("\n", "").replace("\r", ""));
         if (match != null) {
             text.wrapBy(match.getBefore(), match.getAfter());
         }


AnneTheAgile:
#24434 bug cross ref, maybe this is what I am seeing.

AnneTheAgile:
gosh it is so painful when issues aren't on github; here it is; note it does NOT reference back to this PR , merely the code diff, as far as I can see. 
https://issues.jenkins-ci.org/browse/JENKINS-24434

found by searching ; 
https://issues.jenkins-ci.org/secure/IssueNavigator.jspa?mode=hide&reset=true&jqlQuery=project+%3D+JENKINS+AND+status+in+%28Open%2C+%22In+Progress%22%2C+Reopened%29+AND+component+%3D+%27build-failure-analyzer-plugin%27

note the main plugin page; https://wiki.jenkins-ci.org/display/JENKINS/Build+Failure+Analyzer

daniel-beck:
@AnneTheAgile It is not clear what you're trying to say with your comment. A few, hopefully helpful, remarks:

The first issue comment links to a05085cf26d2a901b7ab3a64e8c7b3942313d0f7, which (in the list of branches containing the commit) links to this PR. This PR, in turn, identifies the issue it's fixing in the commit message and PR title.

That query you mention is also linked from the plugin's wiki page as *Open Issues*, so it's rather easily accessible.

Note that in the case of Jenkins, 

* Jira long predates use of Github (which means the entire project issue history is there), 
* hosting at Github is not mandatory (so we need *something* for plugins hosted where there's no integrated tracker, like the old SVN repo that's still a supported means to host plugin code) 
* and using different issue trackers just makes e.g. assigning different or additional components (plugins) to an issue more difficult.

All of which are reasons for using the central Jenkins issue tracker, and against e.g. Github's per-repo issues.

AnneTheAgile:
@daniel-beck , thank you for your kind reply! I apologize my note does not seem to be very clear now that I re-look at it. 

I came here from googling for why my plugin usage is not working right.

What I meant by my remark is that there should be hyperlinks between the two items. It is great that the plugin page itself lists this and the tracker, but from here I should easily be able to get back to them as well. I submitted PR#30 here to add those URLS to the readme to make it trivial.

I am looking forward to using this but right now I can't get it working. I used to do alot of pareto failure reports, so this will be one of my new favorite tools once I get it going. I know a bit of java, so if I can help I'd be happy to.

Thank you for all your work!

AnneTheAgile:
ps it looks like since it is merged it could be closed?

PR #35, opened 2015-03-30
JENKINS-25751 Add to build log indication if a known cause was found

karisivo:


rsandell:
Should the build log really be polluted with the details on what was found?

I think it could be enough to do what findbugs and checkstyle is doing with just a number, e.g. "Found 3 failure causes"

karisivo:
I don't think this will pollute log. For reason why I need this log indication, is that I can use Naginator plug in to trigger rebuild if certain failure is happened. But I don't like to trigger all failures, so I like to see from log what failures is found. Now I think that printing name is maybe not enough, but maybe category should be printed.

We have CI_problems and TestCaseFailure categories, and those CI_Problems may work after rebuild. TestCase failures mostly requires code fix, and no rebuild is wasting time.

I am not sure if it is really problem that one line for each indication is printed to log, if it give needed information.

I will add also category info to print.

karisivo:
Here is log examples

Extra lines added to downstream job:

Build step 'Execute shell' marked build as failure
[BFA] Scanning build for known causes...
**[BFA] Found failure cause(s):**
**[BFA] test from category CI_Problem**
**[BFA] compile failure from category compile_failure**
[BFA] Done. 0s
Finished: FAILURE

[BFA] (.*) from category (.*) 
where \1 is failure cause name and \2 category

Extra lines added to upstream job:
[BFA] Scanning build for known causes...
[BFA] No failure causes found
[BFA] Done. 0s
**[BFA] Found downstream Failure causes ...**
**[BFA] See Downstream_job1#6**
**[BFA] test from category CI_Problem**
**[BFA] compile failure from category compile_failure**
Finished: FAILURE

Same syntax also in upstream jos as in downstream job. Only different is that print is after build failure cause scanning and there is info print that downstream failure cause check is done.

karisivo:
Hi,

could you help me to get this build passed with tests. I fixed two cases which was caused by my commit, but there is two other cases which I cannot reproduce on my machine

 com.sonyericsson.jenkins.plugins.bfa.db.EmbeddedMongoStatisticsTest.testGetFailureCausesPerTimeDayPeriod	1.949	2
 com.sonyericsson.jenkins.plugins.bfa.db.EmbeddedMongoStatisticsTest.testGetUnknownFailureCauseQuotaPerTimeOneDay

And those test cases seems to fail occasional.

karisivo:
Hi,

could you help me to get this build passed with tests. I fixed two cases which was caused by my commit, but there is two other cases which I cannot reproduce on my machine

 com.sonyericsson.jenkins.plugins.bfa.db.EmbeddedMongoStatisticsTest.testGetFailureCausesPerTimeDayPeriod	1.949	2
 com.sonyericsson.jenkins.plugins.bfa.db.EmbeddedMongoStatisticsTest.testGetUnknownFailureCauseQuotaPerTimeOneDay

And those test cases seems to fail occasional.

karisivo:
Yehaa, all is well :)

So when this is available in plugin update center?

PR #36, opened 2015-04-29
Add support for the folders plugin

s0undt3ch:


s0undt3ch:
This initial build will fail because of the jenkins version bump.
I just want to see what's broken with this update and if my changes add any new failures.

s0undt3ch:
The actual folders support will come next

s0undt3ch:
No new failures were added from my code addition.

The test failures are related to the minimum jenkins version bump. Unfortunately, I'm not a java developer and I'm really not familiar with Java or even Jenkins unit testing framework to fix the failing issues. Help from someone with a depth knowledge of it would be greatly appreciated in order to get this into the next release.

I am using my code changes in a Jenkins instance and it's working as supposed.

rsandell:
When I started to dig into fixing the unit tests I noticed that some of ```BuildLogIndicationTest``` is failing because your algorithm doesn't handle when jenkins is running in a different context root, like it does when running the tests on newer versions of core.

s0undt3ch:
So, it fails when Jenkins is served from `/<whatever` right?

rsandell:
Yes,
I started to make some fixes, by first bumping the dependency to 1.580.1 and there I noticed that when the test ran jenkins was running on http://localhost:54545/jenkins/ so ```urlParts[0]``` became ```/jenkins``` so ```fullFolderName``` became ```//jenkins``` and that made ```getItemInstance``` become ```null``` and an NPE then got thrown further down.

s0undt3ch:
Want me to try and address this issue in this PR, or are you intending on pushing your fixed tests with the bumped Jenkins version to the master branch(or any other branch) and I'll rebase and fix?

rsandell:
I stopped once I noticed that the algorithm was wrong, so I've only done pom changes and checkstyle fixes so far. But I can push the few things that I've done and you can pull that to your for if you want to.

rsandell:
Here: https://github.com/jenkinsci/build-failure-analyzer-plugin/tree/s0undt3ch-features/folders-support

s0undt3ch:
I think that I've taken care of the existance(or not) of a URL prefix... Let's see what the tests say....

rsandell:
Seems like it needs some more tweaking ;)

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/Indication.java
```diff
@@ -57,7 +56,6 @@
     /**
      * @param pattern the String value.
      */
-    @DataBoundConstructor
```
rsandell: why?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/dbf/BuildFlowDBF.java
```diff
@@ -1,79 +1,81 @@
...
-                // ignore
-            }
-        }
-
-        result = result.subList(1, result.size());
-
-        return result;
-    }
-}
+/*
```
rsandell: This file looks unchanged

s0undt3ch: It's been quite a while but I believe it as a tabs to spaces change... since the other files use spaces...

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/Indication.java
s0undt3ch: I no longer remember why... I can get that back in if desirable...

s0undt3ch:
As soon as the Jenkins version is bumped the failures start happening. I tried cherry picking your LTS commits into master to confirm this, and, the master branch with the Jenkins version bump failures seem exactly the same as those in this PR.

I am, by no means, a Java developer, so, much of this code was done by try and failing and reading other peoples code.

If I could get a holding a branch where the Jenkins version is bumped and the tests fail, that would be way more simple for me to apply this PR against.... and know if the failures are from my code or not....

s0undt3ch:
Updated my previous comment(it was written on mobile and had some bad phrases in there)...

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/BuildLogIndication.java
```diff
@@ -255,7 +256,30 @@ public FormValidation doMatchText(
...
+                        /*
+                         * We matched a folders job. Let's get the jobs up to the part were the next
+                         * iteration can be continued from
+                         */
+                        String fullFolderName = "";
+                        /* The interestingJobParts string created below is meant to discard everything
+                         * that comes before the first '/job' occurrent which is either nothing or the
+                         * prefix from where jenkins is served, ie: http://localhost/jenkins/job/<job>/<buildNumber>
+                         */
+                        String interestingJobParts = urlParts[0].split("/job", 2)[-1];
```
rsandell: `-1` is not a valid array index

rsandell:
I have managed to fix most of the tests to some degree. What I found so far is that the failures in `MultilineBuildLogIndicationTest` is due to your changes, see inline comment above.

rsandell:
I fixed all other test failures except those failed to your change on the same branch as before

https://github.com/jenkinsci/build-failure-analyzer-plugin/tree/s0undt3ch-features/folders-support

s0undt3ch:
The reason for the data bound question...
```
/home/vampas/projects/jenkins/build-failure-analyzer-plugin/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/Indication.java:[62,11] error: @DataBoundConstructor may not be used on an abstract class (only on concrete subclasses)
```

s0undt3ch:
I've cherry-picked your changes into [https://github.com/s0undt3ch/build-failure-analyzer-plugin/tree/features/bump-to-lts]. As you mentioned, the tests pass. Thank You!

I'll apply my feature on top of that.

rsandell:
Yes a simple look in the git history revealed that I was the one who removed it :)

You should be able to simply pull the branch instead if cherry picking, I pulled from yours and just added my commit. A cherry pick would duplicate that commit which might be unnessesary.

s0undt3ch:
Updated some code.....

Dam! It still fails...

s0undt3ch:
The tests now pass but this is still not correct...

s0undt3ch:
The mongo tests don't fail locally. Closing and reopening to trigger a new build(if this is how it works)...

s0undt3ch:
Ok, I have no idea why the mongo tests are failing(they aren't locally)... does this makes any sense to you @rsandell?

rsandell:
It could be a timezone issue. It's a bit cumbersome for me to pull down your changes since you did a cherry pick instead of a merge/pull, but I'll muster through it ;)

s0undt3ch:
@rsandell Sorry.... Want me to to a clean PR?

rsandell:
Nope, no worries. I got it running locally. It seems to be a jdk 8 vs jdk 7 issue.

Thanks for all your work!

s0undt3ch:
AWESOME!!!!

:fireworks:

PR #39, opened 2015-12-21
Prevent breakage involving build-flow plugin

scoheb:
In certain cases, the Build Flow plugin can return a CancellationExcpetion
when iterating over downstream builds.

This has the effect of breaking both the Job and Build UIs.

This PR protects against this and [FIXES JENKINS-25396]

It is also related to JENKINS-32144 and JENKINS-25092


scoheb:
Any news on this PR?

rsandell:
I've been on vacation, going through it now.

PR #41, opened 2016-02-03
Set the name of the current threadpool worker

rsandell:
Set the name of the current thread in the thread pool to
the cause and indication currently being scanned for
to get some more useful information out of a running Jenkins.


rsandell:
@reviewbybees

oleg-nenashev:
Do we need to restore the original name of the thread when the operation completes? I'd guess so

rsandell:
@oleg-nenashev I'm not so sure it's needed since the ThreadPoolExecutor is shut down directly after all scans are complete.

stephenc:
:bee:

mslusarczyk:
:bee:

PR #49, opened 2016-02-24
Code quality fix - Dead stores should be removed.

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1854 - Dead stores should be removed. 
You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S1854

Please let me know if you have any questions.

Faisal Hameed

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/StatisticsLogger.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/StatisticsLogger.java
@@ -126,7 +126,7 @@ public final class StatisticsLogger {
             Node node = build.getBuiltOn();
             String nodeName = node.getNodeName();
             int timeZoneOffset = TimeZone.getDefault().getRawOffset();
-            String master = "";
+            String master;
 
             String result = build.getResult().toString();
             List<FailureCauseStatistics> failureCauseStatistics = new LinkedList<FailureCauseStatistics>();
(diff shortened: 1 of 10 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/StatisticsLogger.java
```diff
@@ -126,7 +126,7 @@ public void run() {
             Node node = build.getBuiltOn();
             String nodeName = node.getNodeName();
             int timeZoneOffset = TimeZone.getDefault().getRawOffset();
-            String master = "";
+            String master;
```
rsandell: :-1: This will initialize it to `null` instead of an empty string.

faisal-hameed: Initializing with empty string is useless here because in line 140 it will be initialized.
`master = BfaUtils.getMasterName();`

rsandell:
needs a rebase

faisal-hameed:
rebase done.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/StatisticsLogger.java
rsandell: OK
I normally always intialize with the empty string in case of future expansions (I've had too many `nullSometext` printouts in the past). But if it makes you feel better we can have it like this as well ;)

zionyx: It's still a good practice to have the variable set to something over uninitialized. Who knows what will happen to the codes at line 140 in the future and may eventually turn out to produce uninitialized variable exception.

I would vote against the new code.

rsandell: Well at the same time, my IDE usually warns against such things with `Warning: var "master" may not have been initialized`. But it is far often ignored is my guess.

PR #51, opened 2016-03-30
Add option to ignore aborted jobs

Jimilian:
If job was aborted (manually), it doesn't make any sense to analyze results and spend CPU time on it :)
Now it's possible to ignore aborted jobs as well as succeed ones.


src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -504,6 +512,7 @@ public KnowledgeBase getKnowledgeBase() {
     public void configure(StaplerRequest req, JSONObject o) throws Descriptor.FormException, IOException {
         noCausesMessage = o.getString("noCausesMessage");
         globalEnabled = o.getBoolean("globalEnabled");
+        doNotAnalyzeAbortedJob = getBooleanWithDefault(o, "doNotAnalyzeAbortedJob", Boolean.FALSE);
```
rsandell: you don't need a `Boolean` object here since you are never using the `null` value.

And for that you have `o.optBoolean(key, defaultValue)`

src/main/java/com/sonyericsson/jenkins/plugins/bfa/ResultFilter.java
```diff
@@ -0,0 +1,20 @@
+package com.sonyericsson.jenkins.plugins.bfa;
+
+import hudson.model.Result;
+
+public class ResultFilter {
```
rsandell: `shouldScan` is already in `PluginImpl` so having `needToAnalyze` in there is better than a completely new class for just one method

src/main/java/com/sonyericsson/jenkins/plugins/bfa/ResultFilter.java
```diff
@@ -0,0 +1,20 @@
...
+    private static boolean doNotAnalyzeAbortedJobs = false;
+
+    public static void setDoNotAnalyzeAbortedJobs(boolean doNotAnalyzeAbortedJobs) {
+        ResultFilter.doNotAnalyzeAbortedJobs = doNotAnalyzeAbortedJobs;
+    }
+
+    public static boolean needToAnalyze(Result result)  {
+        if (doNotAnalyzeAbortedJobs) {
+            return result != Result.SUCCESS && result != Result.ABORTED;
```
rsandell: `Result.ABORTED` can also mean a timeout and not always a manual cancel. You need to check the cancel cause as well.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/sod/ScanOnDemandTask.java
```diff
@@ -66,14 +66,16 @@ public void run() {
...
-                            && run.getResult().isWorseThan(Result.SUCCESS)
+                            && ResultFilter.needToAnalyze(run.getResult())
                             && run.getNumber() == build.getNumber()) {
                         scanBuild(run);
                     }
                 }
                 endMatrixBuildScan();
             } else {
-                scanBuild(build);
+                if (ResultFilter.needToAnalyze(build.getResult())) {
```
rsandell: This is for a manual scan and should not concern itself with whether it should scan it or not, the user has explicitly asked for the build to be scanned so it should be.

src/test/java/com/sonyericsson/jenkins/plugins/bfa/PluginImplHudsonTest.java
```diff
@@ -40,23 +40,33 @@
...
 /**
- * {@link HudsonTestCase}s for {@link PluginImpl}.
+ * TestCases for {@link PluginImpl}.
  *
  * @author Robert Sandell &lt;robert.sandell@sonyericsson.com&gt;
  */
-public class PluginImplHudsonTest extends HudsonTestCase {
+public class PluginImplHudsonTest {
+    @Rule
+    private JenkinsRule jenkins = new JenkinsRule();
```
rsandell: Example of how to solve the checkstyle public problem https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/src/test/java/com/sonyericsson/hudson/plugins/gerrit/trigger/spec/ParameterModeJenkinsTest.java#L53

src/main/java/com/sonyericsson/jenkins/plugins/bfa/ResultFilter.java
Jimilian: Unfortunately, I didn't find a way to do it properly. So, right now we are marking all "timeouted" jobs as failed.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
Jimilian: Fixed.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/ResultFilter.java
Jimilian: Fixed.

src/test/java/com/sonyericsson/jenkins/plugins/bfa/PluginImplHudsonTest.java
Jimilian: Fixed.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -115,6 +116,7 @@
     private String noCausesMessage;
 
     private Boolean globalEnabled;
+    private Boolean doNotAnalyzeAbortedJob;
```
rsandell: no need for a **B**oolean object here, should just be a **b**oolean type since it defaults to `false`.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -326,6 +333,15 @@ public boolean isGlobalEnabled() {
     }
 
     /**
+     * If this feature is enabled or not. When on all aborted builds will be ignored.
+     *
+     * @return true if on.
+     */
+    public Boolean doNotAnalyzeAbortedJob() {
```
rsandell: please use standard Java bean naming conventions, i.e. `isDoNotAnalyzeAbortedJob`

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -437,6 +462,20 @@ public void setNrOfScanThreads(int nrOfScanThreads) {
     }
 
     /**
+     * Checks if the build with certain result should be analyzed or not.
+     *
+     * @param result the result
+     * @return true if it should be analyzed.
+     */
+    public static boolean needToAnalyze(Result result)  {
+        if (getInstance().doNotAnalyzeAbortedJob()) {
```
rsandell: You can find out if it was manually aborted with the following pseudo code

```java
InterruptedBuildAction a = build.getAction(InterruptedBuildAction.class);
for(CauseOfInterruption cause : a.getCauses()) {
    if (cause instanceof UserInterruption) {
        //User aborted the run, now do something
    }
}
```

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
```diff
@@ -111,5 +111,9 @@
                            default="${it.sodVariables.DEFAULT_SOD_COREPOOL_THREADS}"/>
             </f:entry>
         </f:section>
+        <f:entry title="${%Do not analyze aborted jobs}"
```
rsandell: normally the title is shorter than the description :) So the description does not add any value to the user being there. Either switch them or remove the description.

rsandell:
Tis is starting to look really good.
A help file for the new field would be even better.

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
Jimilian: :) ~~Done~~ Removed

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
Jimilian: Done

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
Jimilian: Done

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
Jimilian: Thanks you. It works, but I decide to avoid this check. But I added information how to handle case with aborted jobs to help.
Main point why I don't want to add this check - because I wish to have black list (all aborted jobs except time-outed) instead of white (only aborted by user). But I can't implement it, because I don't want to add dependency for "timeout" plugin.

Jimilian:
Hi! 
Is it ok now?

rsandell:
@Jimilian sorry, I've been on a short vacation. I'll take a look asap.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
rsandell: OK, if you want to add it in the future you can look at the class name of the cause to avoid adding the dependency. e.g.
```java
if (cause.getClass().getName().equals("org.jenkinsci.plugins.WhateverTheNameNowIs"))
```
A bit ugly but should work ;)

PR #52, opened 2016-04-15
Pipeline support for Build Failure Analyzer

TWestling:
With this change, Pipeline projects are analyzed
by the BFA. The major change was changing AbstractBuild
to Run and AbstractProject to Job, which would also make
other project and build types that aren't AbstractBuild/Project
start working with the BFA.

Some issues remain and should be fixed in future changes:
Downstream builds linking. If a Pipeline project starts downstream builds
which fail, these won't be shown on the Pipeline build page.

"Built on". In the statistics logging, adding which node the Pipeline
run was run on doesn't work. A Pipeline project could use many nodes
and a new structure for this in the statistics must be created.

TokenMacro plugin integration. The TokenMacro evaluation uses AbstractBuild.
This needs to be changed to Run in the TokenMacro plugin in order for the
evaluation to work with Pipeline runs.

Change-Id: I2a0debed0a13c051d3614c5ceece77a31232ac3d

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/KnowledgeBase.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/KnowledgeBase.java
@@ -30,6 +30,8 @@ import com.sonyericsson.jenkins.plugins.bfa.model.FailureCause;
 import com.sonyericsson.jenkins.plugins.bfa.statistics.Statistics;
 import com.sonyericsson.jenkins.plugins.bfa.utils.ObjectCountPair;
 import hudson.ExtensionList;
+import hudson.Util;
+import hudson.model.AbstractBuild;
 import hudson.model.Describable;
 import hudson.model.Descriptor;
 import hudson.model.Run;
@@ -351,8 +353,28 @@ public abstract class KnowledgeBase implements Describable<KnowledgeBase>, Seria
      * Removes the build failure cause of particular build.
      * @param build the build.
      * @throws Exception if something in the KnowledgeBase handling goes wrong.
+     * @deprecated use {@link #removeBuildfailurecause(hudson.model.Run)}
      */
-    public abstract void removeBuildfailurecause(Run build) throws Exception;
+    @Deprecated
+    public void removeBuildfailurecause(AbstractBuild build) throws Exception {
+        if (Util.isOverridden(KnowledgeBase.class, getClass(), "removeBuildfailurecause", Run.class)) {
+            removeBuildfailurecause((Run)build);
+        }
+    }
+
+    /**
+     * Removes the build failure cause of particular build.
+     * @param build the build.
+     * @throws Exception if something in the KnowledgeBase handling goes wrong.
+     */
+    public void removeBuildfailurecause(Run build) throws Exception {
+        if (Util.isOverridden(KnowledgeBase.class, getClass(), "removeBuildfailurecause", AbstractBuild.class)) {
+            removeBuildfailurecause((AbstractBuild)build);
+        }
+    }
+
+
+
     /**
      * Descriptor for {@link KnowledgeBase}s.
      */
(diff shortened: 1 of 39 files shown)


TWestling:
Adding tests for WorkFlowRuns seems to require upgrading the jenkins core dependency in BFA to 1.642.3, which the Pipeline plugin requires. Since Pipeline isn't even a dependency for BFA, this seemed like overkill.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/KnowledgeBase.java
```diff
@@ -349,10 +349,10 @@ public Date getCreationDateForCause(String id) {
 
     /**
      * Removes the build failure cause of particular build.
-     * @param build the AbstractBuild.
+     * @param build the build.
      * @throws Exception if something in the KnowledgeBase handling goes wrong.
      */
-    public abstract void removeBuildfailurecause(AbstractBuild build) throws Exception;
+    public abstract void removeBuildfailurecause(Run build) throws Exception;
```
rsandell: This will probably break backwards compatibility for other KnowledgeBase implementations.

I think you need to add a new method that takes a Run and deprecate the old one and make both non abstract.
There are an example or two in the gerrit trigger I think.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureReader.java
```diff
@@ -74,15 +74,15 @@ public FailureReader(Indication indication) {
      * is found in the log of the given build; return null otherwise.
      * @throws IOException if so.
      */
-    public abstract FoundIndication scan(AbstractBuild build) throws IOException;
+    public abstract FoundIndication scan(Run build) throws IOException;
```
rsandell: Same here, although we could take the risk tha no one else has implemented the extension point, but we all know what happens to assumptions ;)

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/ScannerOffJobProperty.java
```diff
@@ -36,7 +36,7 @@
  * @deprecated {@link ScannerJobProperty} is used instead, but this is kept to be able to de-serialize old jobs.
  */
 @Deprecated
-public class ScannerOffJobProperty extends JobProperty<AbstractProject<?, ?>> implements Serializable {
+public class ScannerOffJobProperty extends JobProperty<Job<?, ?>> implements Serializable {
```
rsandell: You didn't need to change this one, it only exists for older jobs and is deserialized as a `ScannerJobProperty`

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/dbf/DownstreamBuildFinder.java
```diff
@@ -49,17 +49,17 @@
...
+    protected static final List<Run<?, ?>> EMPTY =
+            Collections.unmodifiableList(new LinkedList<Run<?, ?>>());
 
     /**
      * Return a list of all downstream builds originating from provided build.
      *
      * @param build get the downstream build(s) relative this build
      * @return a list with downstream builds
      */
-    public abstract List<AbstractBuild<?, ?>> getDownstreamBuilds(
```
rsandell: same backcompat problem here.

rsandell:
If you want to add a pipeline test you can add the dependency in the test scope, the pipeline plugin will then be available during hpi:run and during unit tests but will not be a runtime dependency when released.

There are older versions of the pipeline plugin that works just fine that has older core dependencies, like the one used in gerrit trigger ;)

TWestling:
Good input, will fix.
Didn't think of trying an older pipeline plugin in test, I'll do that.

TWestling:
The test that have errors run fine on my machine, looks like time difference is causing some issues.

rsandell:
:+1:

PR #53, opened 2016-08-08
Add limit for log size to be scanned

Jimilian:
BFA plugin tries to parse files in parallel (one thread per `pattern`), so, sometimes we had a lot of problems then one job produced huge log for some reason (i.e. printing some stats in some `while(true)` loops) and all workers are busy, and all other jobs are waiting them. To avoid such cases I introduced limitation for log size. 
I suppose that `onDemand` should work without checking this limit.

Default value is 1 GB. (But I think optimal value should be between 20 and 50 MB).


Jimilian:
Now default value is 0. It means that default behaviour stays as before.

@rsandell if you are ok with this PR, maybe you can make a small promo for this option in "Tips & Tricks" area? :)

rsandell:
@Jimilian thanks!

The wiki is open for anyone with an account to edit, so anyone is free to contribute to the tips & tricks section ;)

emoshaya-cognito:
Hi there what is the value field format? is it bytes or megabytes? i.e. if I want to set the "Max size of log file" to 500mb, do I put the value in bytes or just 500?

Jimilian:
@emoshaya-cognito it's megabytes. And it's better to set this limit to smaller value - 50 or even 10 MB. Also big logs are big problem for Jenkins.

PR #55, opened 2016-08-30
Allow formatted causes to be used in pipelines

pauxus:
This PR splits the Token into Token and Renderer. Aside from SOC reasons, this also allows the Renderer to be used separately, in my use case inside of a pipeline (which does not currently support TokenMacros).

I.e. one could now do something like:

```groovy
BuildFailureScanner.scan(currentBuild.rawBuild, new PrintStream(new NullStream()))
String causes = new Renderer(useHtmlFormat: true, includeTitle: false).render(currentBuild.rawBuild.getAction(FailureCauseBuildAction.class))
```

The causes String can than be included in generated mails.

Note that the next step should be to wrap the renderer into a nice dsl keyword.


rsandell:
Well this split seems OK, but I'm not sure I'm liking what you are planning next.
As you can't run the code you describe above in a sandboxed pipeline, it is using unsafe code.

A more dsl like approach would be something like

```groovy
def foundFailures = withBfa {
 //my steps and all
}
```
or perhaps better
```groovy
withBfa block: {
  //my steps and all
}, whenFound: { def foundFailures ->
  //do something with foundFailures like sending email or
  String report = bfa.render(failures: foundFailures, useHtmlFormat: true, includeTitle: false)
}
```

where `foundFailures` is a list of some data structure that GroovyCPS can handle (i.e. Serializable and whitelisted getters)

pauxus:
Agreed about GroovyCPS whitelists. The problem with the block is that BFA runs around every pipeline anyway (although to late). I think introducing another block would make the resulting pipeline only harder to read.

Consider the use case of a complex pipeline which sends different error mails for different stages. In that case you would already wrap code parts in try / finally blocks. In order for that to work, the withBfa method would also need to approach the whenFound block inside a finally block.

What about:

```groovy
try {
  //my steps and all
} catch (e) { 
  def foundFailures = bfa.scan()
  //do something with foundFailures like sending email or
  String report = bfa.render(failures: foundFailures, useHtmlFormat: true, includeTitle: false)
}
```

rsandell:
Well, in my example the withBfa would be equivalent of a try/finally block, so you wouldn't need the latter ;)
I'm not sure it is possible to do though, but I believe it is.

jglick:
My $0.02: do not bother with blocks, and certainly do not do anything which requires the user to make (trusted) Java calls like the original PR description suggested. Just define an `AbstractSynchronousStepExecution<String>`:

```groovy
try {
  node {
    sh 'make world'
  }
} catch (e) {
  def report = bfa useHtmlFormat: true, includeTitle: false
  mail to: …, body: report
  throw e
}
```

and leave it to the script author to decide when and how to use it.

rsandell:
OK, and the bfa step could add a marker Action or something to the build that tells the normal scanning to not scan this build, so we don't get double scanning.

pauxus:
Sounds good. This PR was designed as a step in that direction anyway. 

@rsandell Should I create a separate PR for the StepExecution or should I include it simply in this one?

pauxus:
Ok, I tried implementing this, but there currently is one problem: The logs for WorkflowRuns are splitted and only combined when the workflow is finished.

What we would need to make this work would be a change to WorkflowRun to allow to get partial logs... I don't know if this is really getting us anywhere...

rsandell:
There is most likely some kind of live combine going on somewhere, otherwise we wouldn't get a live log on the pipeline run page.

PR #57, opened 2016-09-06
Use fixed pool size, parse all single line matchers in one thread

Jimilian:
So, this is ambitious PR:
- Set language level to Java 7
- Thread pool is constant now - doesn't depend on number of failed job
- All single line patterns are parsed in one thread - reduces CPU load and memory pressure A LOT

I understand that right now this PR is far away from final state. But it already works as expected.
My questions are:
- Can I keep Java 7?
- Can I delete\rework scan(Run build, PrintStream buildLog)?
- Can I rework chain Cause->Indication->Pattern? I don't like current solution `setter`+`getter`, but I didn't want to make significant changes without asking it here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -230,8 +230,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.7</source>
-                    <target>1.7</target>
+                    <source>1.6</source>
+                    <target>1.6</target>
                 </configuration>
             </plugin>
             <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -230,8 +230,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.7</source>
+                    <target>1.7</target>
                 </configuration>
             </plugin>
             <plugin>
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
@@ -197,7 +197,7 @@ public class BuildFailureScanner extends RunListener<Run> {
      */
     private static List<FoundFailureCause> findCauses(final Collection<FailureCause> causes,
                                                       final Run build, final PrintStream buildLog) {
-        final List<FoundFailureCause> foundFailureCauseList = new ArrayList<>();
+        final List<FoundFailureCause> foundFailureCauseList = new ArrayList<FoundFailureCause>();
         long start = System.currentTimeMillis();
 
         THREAD_POOL_EXECUTOR.setCorePoolSize(PluginImpl.getInstance().getNrOfScanThreads());
@@ -215,7 +215,9 @@ public class BuildFailureScanner extends RunListener<Run> {
 
         try {
             THREAD_POOL_EXECUTOR.submit(runnable).get();
-        } catch (InterruptedException | ExecutionException e) {
+        } catch (InterruptedException e) {
+            buildLog.print("[BFA] was interrupted: " + e);
+        } catch (ExecutionException e) {
             buildLog.print("[BFA] was interrupted: " + e);
         }
 
@@ -258,10 +260,10 @@ public class BuildFailureScanner extends RunListener<Run> {
     private static List<FoundFailureCause> findIndications(final Collection<FailureCause> causes,
                                                          Run build,
                                                          PrintStream buildLog) {
-        List<FoundFailureCause> foundFailureCauses = new ArrayList<>();
+        List<FoundFailureCause> foundFailureCauses = new ArrayList<FoundFailureCause>();
 
-        List<FailureCause> singleLineCauses = new ArrayList<>();
-        List<FailureCause> notOnlySingleLineCauses = new ArrayList<>();
+        List<FailureCause> singleLineCauses = new ArrayList<FailureCause>();
+        List<FailureCause> notOnlySingleLineCauses = new ArrayList<FailureCause>();
 
         for (FailureCause cause : causes) {
             boolean atLeast = false;
@@ -278,47 +280,53 @@ public class BuildFailureScanner extends RunListener<Run> {
             }
         }
 
+        BufferedReader reader = null;
         try {
-            try (BufferedReader reader = new BufferedReader(build.getLogReader())) {
-                foundFailureCauses.addAll(
-                        FailureReader.scanSingleLinePatterns(
-                                singleLineCauses,
-                                build,
-                                reader,
-                                build.getLogFile().getName()));
-            }
-
-            for (FailureCause cause : notOnlySingleLineCauses) {
-                List<FoundIndication> foundIndications = new ArrayList<>();
-                for (Indication indication : cause.getIndications()) {
-                    long start = System.currentTimeMillis();
-
-                    Thread.currentThread().setName("BFA-scanner-"
-                            + build.getFullDisplayName() + ": "
-                            + cause.getName() + "-"
-                            + indication.getUserProvidedExpression());
-
-                    final FoundIndication foundIndication = findIndication(indication, build, buildLog);
-                    if (foundIndication != null) {
-                        foundIndications.add(foundIndication);
-
-                        if (logger.isLoggable(Level.FINER)) {
-                            logger.log(Level.FINER, "[BFA] [{0}] [{1}] {2}ms", new Object[]{build.getFullDisplayName(),
-                                    cause.getName(),
-                                    String.valueOf(System.currentTimeMillis() - start), });
-                        }
-                    }
-                }
-
-                foundFailureCauses.add(new FoundFailureCause(cause, foundIndications));
-            }
-
-            return foundFailureCauses;
+            reader = new BufferedReader(build.getLogReader());
+            foundFailureCauses.addAll(
+                    FailureReader.scanSingleLinePatterns(
+                            singleLineCauses,
+                            build,
+                            reader,
+                            build.getLogFile().getName()));
         } catch (IOException e) {
             buildLog.print("[BFA] Exception during parsing file: " + e);
+        } finally {
+            if (reader != null) {
+                try {
+                    reader.close();
+                } catch (IOException e) {
+                    logger.log(Level.WARNING, "Failed to close the reader. ", e);
+                }
+            }
         }
 
-        return new ArrayList<>();
+        for (FailureCause cause : notOnlySingleLineCauses) {
+            List<FoundIndication> foundIndications = new ArrayList<FoundIndication>();
+            for (Indication indication : cause.getIndications()) {
+                long start = System.currentTimeMillis();
+
+                Thread.currentThread().setName("BFA-scanner-"
+                        + build.getFullDisplayName() + ": "
+                        + cause.getName() + "-"
+                        + indication.getUserProvidedExpression());
+
+                final FoundIndication foundIndication = findIndication(indication, build, buildLog);
+                if (foundIndication != null) {
+                    foundIndications.add(foundIndication);
+
+                    if (logger.isLoggable(Level.FINER)) {
+                        logger.log(Level.FINER, "[BFA] [{0}] [{1}] {2}ms", new Object[]{build.getFullDisplayName(),
+                                cause.getName(),
+                                String.valueOf(System.currentTimeMillis() - start), });
+                    }
+                }
+            }
+
+            foundFailureCauses.add(new FoundFailureCause(cause, foundIndications));
+        }
+
+        return foundFailureCauses;
     }
 
     /**
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
@@ -301,10 +301,9 @@ public class BuildFailureScanner extends RunListener<Run> {
                     foundIndications.add(findIndication(indication, build, buildLog));
 
                     if (logger.isLoggable(Level.FINER)) {
-                        logger.log(Level.FINER, "[BFA] [{0}] [{1}] {2}ms", new Object[]
-                                {build.getFullDisplayName(),
-                                        cause.getName(),
-                                        String.valueOf(System.currentTimeMillis() - start),});
+                        logger.log(Level.FINER, "[BFA] [{0}] [{1}] {2}ms", new Object[]{build.getFullDisplayName(),
+                                cause.getName(),
+                                String.valueOf(System.currentTimeMillis() - start), });
                     }
                 }
 
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
@@ -29,12 +29,13 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.util.ArrayList;
-import java.util.List;
-import java.util.LinkedList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
@@ -61,6 +62,8 @@ import hudson.model.TaskListener;
 import hudson.model.listeners.RunListener;
 import hudson.tasks.test.AbstractTestResultAction;
 import hudson.tasks.test.TestResult;
+
+import javax.annotation.Nonnull;
 
 /**
  * Looks for Indications, trying to find the Cause of a problem.
@@ -94,7 +97,7 @@ public class BuildFailureScanner extends RunListener<Run> {
     }
 
     @Override
-    public void onCompleted(Run build, TaskListener listener) {
+    public void onCompleted(Run build, @Nonnull TaskListener listener) {
         logger.entering(getClass().getName(), "onCompleted");
 
         if (PluginImpl.isSizeInLimit(build)) {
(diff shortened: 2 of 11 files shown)


Jimilian:
Close and reopen to re-trigger. (Jenkins report check style issues on empty lines)

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/BuildLogFailureReader.java
```diff
@@ -61,46 +62,23 @@ public BuildLogFailureReader(BuildLogIndication indication) {
...
-     * Scans for indications of a failure cause in a build log. Note: If an exception
-     * occurs during the scanning, information about the exception is appended to
-     * the build log.
-     *
-     * @param build the build to scan for indications.
-     * @param buildLog the log of the build.
-     * @return a FoundIndication if something was found, null if not.
-     */
-    @Override
-    public FoundIndication scan(Run build, PrintStream buildLog) {
```
Jimilian: Should I keep this method as well?

rsandell:
I'm not sure I understand your motivation for doing this kind of invasive change, to me sacrificing memory footprint for a quicker set of scans is well within reason, hardware is cheap but patient users are not ;)

Anyway I'm going to Jenkins World next week so I won't have time to look at this until I'm back.

Jimilian:
Motivation is pretty simple - I want to be sure that BFA will not kill my Jenkins instance and would be fast enough to provide feedback.
To achieve first goal, I need to make number of threads constant (see picture that can happen in some cases). It allocates a lot of memory (it's 10 seconds snapshot) and run a lot threads ([CPU starvation](https://en.wikipedia.org/wiki/Starvation_(computer_science))).
To achieve second goal, I need to parse as much patterns as possible in one thread. It's quite easy to do, if you have a lot of single line patterns. In this case you can parse all patterns in once.


bfa_rocks

Jimilian:
Close and reopen to re-trigger. (Again statistics tests report wrong date).

Jimilian:
@rsandell I hope that you had recovered already after Jenkins World :)
Do you have a time to take a look into this PR?

I want to mention only one thing that we need to address - `TIMEOUT_FILE`. To keep old behaviour I can multiply it on number of potential causes. Move this value to parameters or just set it in 2 minutes.

Jimilian:
> to me sacrificing memory footprint for a quicker set of scans is well within reason
Just to make it clear. If you have only single line patterns with this PR BFA works much-much faster: 
- there is no need to read file multiple time
- all patterns could be checked at once
- there is no hidden queue (you have only 5 executors and 10 patterns = 5 patterns are waiting free executor)

In case of multiline parser plugin uses constant thread pool size instead of thread pool per build. So, user can specify how much BFA threads he wants: 1, 5 or 10. Depends on machine resources.

dcoraboeuf:
I'm really interested in the outcome of this PR. We have a big volume (> 5000 jobs, logs > 10 Mb, ) and we had to disable BFA because builds were taking too much time to finish.

Jimilian:
@dcoraboeuf , you already can play with this.
It's enough to download hpi artefact from Jenkins: https://jenkins.ci.cloudbees.com/job/plugins/job/build-failure-analyzer-plugin/230/com.sonyericsson.jenkins.plugins.bfa$build-failure-analyzer/artifact/com.sonyericsson.jenkins.plugins.bfa/build-failure-analyzer/1.17.2-SNAPSHOT/build-failure-analyzer-1.17.2-SNAPSHOT.hpi
And install it in "Advanced" panel.

dcoraboeuf:
I'll have to use it in our acceptance environment (I cannot test it in production) but the volume is much lower. And I won't be able to do this for quite some time due to planning constraints :( Thanks all the same, I'll try to follow this.

Jimilian:
@rsandell Any chance to see it in upstream?

oleg-nenashev:
CC @jenkinsci/code-reviewers

mmitche:
I think this is a good PR.  On my instance (very large), we actually can occasionally run out of threads when the BFA is run, which causes cascading issues elsewhere. Specifically, the timeout threads really hurt too.

I'm going to give this is a go.

Jimilian:
@mmitche thx! I was so close to lost hope that even tried to promote this PR: http://www.slideshare.net/jimbrighthead/continuous-integration-on-steroids/59

mmitche:
We have ~30 indications.  There may be other issues, but my instance jumped from ~2k to ~10k threads (and crashed).  this is at least one culprit based on the thread dump.

The issue may be that native resources start to not be collected, and bam bam...done.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -248,34 +247,75 @@ public void run() {
...
+                List<FoundIndication> foundIndications = new ArrayList<>();
+                for (Indication indication : cause.getIndications()) {
+                    long start = System.currentTimeMillis();
+
+                    Thread.currentThread().setName("BFA-scanner-"
+                            + build.getFullDisplayName() + ": "
+                            + cause.getName() + "-"
+                            + indication.getUserProvidedExpression());
+
+                    foundIndications.add(findIndication(indication, build, buildLog));
```
mmitche: findIndication can return null for the multi-line indications (if they don't match), which adds a null indication and causes an NPE down the line.

pom.xml
```diff
@@ -230,8 +230,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.7</source>
+                    <target>1.7</target>
```
oleg-nenashev: I would go even further:
1) Update parent POM to 2.x latest
2) Update Core requirement to 1.609.x or above
3) Get rid of this section, you will get Java 7 by default

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/BuildLogFailureReader.java
oleg-nenashev: I am not sure if deletion of override is required

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -248,34 +247,75 @@ public void run() {
...
+
+                    if (logger.isLoggable(Level.FINER)) {
+                        logger.log(Level.FINER, "[BFA] [{0}] [{1}] {2}ms", new Object[]{build.getFullDisplayName(),
```
oleg-nenashev: Have you checked that the logger really processes this formatter string with double brackets?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -248,34 +247,75 @@ public void run() {
...
+                                String.valueOf(System.currentTimeMillis() - start), });
+                    }
+                }
+
+                foundFailureCauses.add(new FoundFailureCause(cause, foundIndications));
+            }
+
+            return foundFailureCauses;
+        } catch (IOException e) {
+            buildLog.print("[BFA] Exception during parsing file: " + e);
```
oleg-nenashev: Also exception to System log?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/Indication.java
```diff
@@ -53,6 +54,8 @@
      */
     private String pattern;
 
+    private FailureCause cause;
```
oleg-nenashev: Unused field?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
```diff
@@ -58,6 +59,7 @@
     private String pattern;
     private Run build;
     private String matchingString;
+    private FailureCause cause;
```
oleg-nenashev: unused field?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
```diff
@@ -58,6 +59,7 @@
     private String pattern;
     private Run build;
```
mmitche: Is this supposed to be here?  Doesn't this end up creating a cycle that makes collection difficult?

Jimilian: It was added to support Pipeline builds (according `git blame`)

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
Jimilian: Done

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
Jimilian: It was here for ages - I didn't change anything :/

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
Jimilian: Should I add it like this
```
if (logger.isLoggable(Level.FINER)) {
            logger.log(Level.FINER, ... );
```
?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/BuildLogFailureReader.java
Jimilian: Accident :)

pom.xml
Jimilian: I hope that I understood that you wanted.

rsandell:
Sorry, this looked to me like a big change or at least touched some critical paths, and I have been trying to carve out a big enough chunk of time in my day to give the review a thorough look and not just a quick glance. But my schedule has just been too full. I will get to it a some point, unless @TWestling wakes up and does it before me ;)

pom.xml
rsandell: Yes bumping to parent pom 2 would cause too much unrelated changes and make the review even harder/longer ;)

mmitche:
Can confirm this ran in production for quite a while with no issues other than the one I already reviewed.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
```diff
@@ -59,7 +58,6 @@
...
     private String matchingString;
-    private FailureCause cause;
```
Jimilian: @rsandell @oleg-nenashev removing this stuff produces
```
SEVERE: Failed Loading plugin build-failure-analyzer
java.io.IOException: Failed to initialize
	at hudson.ClassicPluginStrategy.load(ClassicPluginStrategy.java:442)
	at hudson.PluginManager$2$1$1.run(PluginManager.java:406)
	at org.jvnet.hudson.reactor.TaskGraphBuilder$TaskImpl.run(TaskGraphBuilder.java:169)
	at org.jvnet.hudson.reactor.Reactor.runTask(Reactor.java:282)
	at jenkins.model.Jenkins$8.runTask(Jenkins.java:926)
	at org.jvnet.hudson.reactor.Reactor$2.run(Reactor.java:210)
	at org.jvnet.hudson.reactor.Reactor$Node.run(Reactor.java:117)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.io.IOException: Unable to read /var/lib/jenkins/build-failure-analyzer.xml
	at hudson.XmlFile.unmarshal(XmlFile.java:165)
	at hudson.Plugin.load(Plugin.java:264)
	at com.sonyericsson.jenkins.plugins.bfa.PluginImpl.start(PluginImpl.java:151)
	at hudson.ClassicPluginStrategy.startPlugin(ClassicPluginStrategy.java:450)
	at hudson.ClassicPluginStrategy.load(ClassicPluginStrategy.java:439)
	... 9 more
Caused by: com.thoughtworks.xstream.converters.ConversionException: null : null
---- Debugging information ----
cause-exception     : java.lang.NullPointerException
cause-message       : null
class               : com.sonyericsson.jenkins.plugins.bfa.PluginImpl
required-type       : com.sonyericsson.jenkins.plugins.bfa.PluginImpl
converter-type      : hudson.util.RobustReflectionConverter
path                : /com.sonyericsson.jenkins.plugins.bfa.PluginImpl
--
java.io.IOException: Unable to load org.jenkinsci.plugins.globalEventsPlugin.GlobalEventsPlugin from groovy-events-listener-plugin
	at hudson.ClassicPluginStrategy.load(ClassicPluginStrategy.java:425)
	at hudson.PluginManager$2$1$1.run(PluginManager.java:406)
	at org.jvnet.hudson.reactor.TaskGraphBuilder$TaskImpl.run(TaskGraphBuilder.java:169)
	at org.jvnet.hudson.reactor.Reactor.runTask(Reactor.java:282)
	at jenkins.model.Jenkins$8.runTask(Jenkins.java:926)
	at org.jvnet.hudson.reactor.Reactor$2.run(Reactor.java:210)
	at org.jvnet.hudson.reactor.Reactor$Node.run(Reactor.java:117)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.ExceptionInInitializerError
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at hudson.ClassicPluginStrategy.load(ClassicPluginStrategy.java:419)
	... 9 more
```
I restored this line (and from `Indication`) and everything works again. So, what should I do? Restore it? And mark `transient`?

rsandell: https://wiki.jenkins-ci.org/display/JENKINS/Hint+on+retaining+backward+compatibility

pom.xml
rsandell: But you should bump code dependency to 1.625.3 to indicate the Java 7 requirement if you actually need it, otherwise stay on Java 6 as that was the Java version supported by Jenkins up until 1.612

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
Jimilian: Done.

pom.xml
Jimilian: Ok, I switched back to Java 6. I wasn't brave enough to fix compilation for all API changes between `1.612` and `1.625.3`.

Jimilian:
@rsandell, @TWestling, what should I do to merge this PR? :)

TWestling:
Sorry, I haven't had the time to look at this thoroughly. I will try to squeeze it in during this week,

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/Indication.java
```diff
@@ -54,6 +55,12 @@
     private String pattern;
 
     /**
+     * @deprecated, kept for backwards compatibility.
+     */
+    @Deprecated
+    private transient FailureCause cause;
+
```
Jimilian: Unfortunately, it's not possible. I tried - but Jenkins fails to deserialize `config.xml` in this case. I don't remember exactly the error, but it was something about "unknown field".

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -190,44 +197,38 @@ public static void scan(Run build, PrintStream buildLog) {
...
-        final List<FoundFailureCause> foundFailureCauseList =
-                Collections.synchronizedList(new LinkedList<FoundFailureCause>());
+        final List<FoundFailureCause> foundFailureCauseList = new ArrayList<FoundFailureCause>();
         long start = System.currentTimeMillis();
-        ThreadPoolExecutor executor = (ThreadPoolExecutor)Executors.
-                newFixedThreadPool(PluginImpl.getInstance().getNrOfScanThreads());
+
+        THREAD_POOL_EXECUTOR.setCorePoolSize(PluginImpl.getInstance().getNrOfScanThreads());
+        THREAD_POOL_EXECUTOR.setMaximumPoolSize(PluginImpl.getInstance().getNrOfScanThreads());
+
```
Jimilian: So, before user configured number of threads used for single build. Now it's global pool - Jenkins/plugin allows to specify maximum number of threads used for parsing.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -190,44 +197,38 @@ public static void scan(Run build, PrintStream buildLog) {
...
-                    Thread.currentThread().setName("BFA-scanner-"
-                                                           + build.getFullDisplayName() + ": "
-                                                           + cause.getName());
-                    List<FoundIndication> foundIndications = findIndications(cause, build, buildLog);
-                    if (!foundIndications.isEmpty()) {
-                        FoundFailureCause foundFailureCause = new FoundFailureCause(cause, foundIndications);
-                        foundFailureCauseList.add(foundFailureCause);
-                    }
+                    Thread.currentThread().setName("BFA-scanner-" + build.getFullDisplayName());
+                    foundFailureCauseList.addAll(findIndications(causes, build, buildLog));
```
Jimilian: It could create delay at first time, but after that user can increase number of threads if server has enough capacity or decrease number of regular expressions (or optimize it :) ).

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
TWestling: Ah, of course, that makes sense.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/Indication.java
TWestling: This doesn't make sense to me. The field only exists in one of your previous commits of this pull request. So your failure to deserialize should only happen when running an instance where you have ran one of your earlier commits, something that won't happen for anyone else.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
TWestling: The way I thought it worked was that you start one thread for all single lines, which is faster than going through the build log several times.

But this one thread is also used for handling everything in the old way, for Causes with at least one non-single-line indication. So in a setup with a lot of non-single-lines, what earlier was done in parallel in a lot of threads, is now done in a serial way, with one thread. 

This doesn't seem to be tweakable in any way for a specific build, only for the server as a whole. If any given log is long and you have lots of MultiLines, it will take a long time, is this correct or am I misreading something?

Jimilian: Yes, you are right. I can add support for this case as well (one task for all pure-single-line regular expressions, and one task per not-only-single-line expression). Should I do it in this PR?

TWestling: Yeah, I think that would make sense, then we aren't sacrificing one use case in favor of another.

TWestling: @Jimilian Is there an ETA for the changes for not-only-single-line expressions? :)

Jimilian: @TWestling Oh, man! One month had passed already! I will try to implement it on next week. But I'm doing it in my spare time...

TWestling: @Jimilian No rush, just thought I'd ping you so that it wasn't forgotten altogether :)

Jimilian:
@rsandell, @TWestling This last commit this PR does:
* One global thread pool for BFA (instead of thread pool per build)
* All single line causes will be parsed together (one shared stream, one thread)
* All non-single-line-only causes  will be parsed in parallel (one stream per regexp, one thread per cause)
* Add several test cases

TWestling:
@Jimilian Sorry about the repeated delays, just as you commited your latest changes, I went on parental leave for a month. I'm back in office today and will take a look at your changes.

Jimilian:
@TWestling, congratulations ;)

TWestling:
@Jimilian Thank you :)

mmitche:
Awesome

PR #59, opened 2016-10-12
MultiLineBuildLogIndication performance and bug fixes

GLundh:
This fork provides some needed MultiLineBuildLogIndication
performance and bug fixes (see commit messages for more details):

**JSON Serialization**

Since inherited classes from Indication does not declare
@JsonProperty("pattern") in their constructors, the JSON
writer will use the default constructor and populate fields
using a matching (by name) getter.

This leads to one particular bad bug, where
MultiLineBuildLogIndication returns a faulty modified pattern
using the @Overidden getPattern() during JSON serialization.

**Performance fixes.**

Test data:
  Log: 49MB log @ 280.000 lines
  Multiline pattern: Simple pattern, matching +20 lines at end of log
  **Pre-fix: 32 seconds** (Note that the hardcoded timeout is 10 seconds)
  **Post-fix: 1.7 seconds**

**Multiline patterns could only match beginning of a line**

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureReader.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureReader.java
@@ -31,6 +31,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Scanner;
 import java.util.logging.Logger;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import com.google.common.base.Joiner;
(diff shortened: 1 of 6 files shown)


rsandell:
At least one test failure seems legit, rebuilding since some of the tests where due to timezone issues.

rsandell:
Yepp, the three remaining test failures seems legit.

GLundh:
Thanks for looking at the code Bobby. I have updated the testcases.

GLundh:
Hang on. Just noticed the checkstyle.

rsandell:
It's hard to tell; does this change require a minor version bump? Or is the behaviour consistent enough with previous versions to just bump the micro version?

GLundh:
Thanks!

> It's hard to tell; does this change require a minor version bump?

Hard to say indeed.

Multiline + Mongo is broken today. So that fix should go in as soon as possible.

The performance stuff should be drop in replacement. But looking at the release history, performance fixes have motivated major releases in the past (see 1.15 -> 1.16). Or perhaps it was the xss-fix that motivated that version jump?

rsandell:
Yes, it was the size of the xss fix that prompted the minor bump instead of a micro bump.

PR #60, opened 2016-10-12
New Extension Point: IndicationFilter

GLundh:
IndicationFilter allows plugins to hide specific indications
from the CauseManagement UI.

Indications added before a Filter is applied will still continue
to run.

The main purpose of this ExtensionPoint is to allow plugin
contributed indications to hide core indications which supply
similiar (or just a subset) of the core indicators' functionality.

Change-Id: I0dc492cdedb1232850262e1d11db35009031d9a3


GLundh:
We don't have a use case for this anymore. If anyone else see a use case, feel free to reopen it.

PR #61, opened 2016-12-05
Send "unknown" to Gerrit if no bfa patterns were found

Jimilian:
Now BFA reports something like "No problems were identified..." if there is no matches. But in case of gerrit feedback it doesn't send anything, so, user should click on feedback and discover failed child job by himself.

I think we can easily simplify this case by providing at least failed build directly to Gerrit. How it would be in Gerrit:

screen shot 2016-12-02 at 16 51 14

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/GerritMessageProviderExtension.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/GerritMessageProviderExtension.java
@@ -98,5 +98,12 @@ public class GerritMessageProviderExtension extends GerritMessageProvider {
             .append(displayData.getLinks().getBuildUrl())
             .append(" )");
         }
+
+        if (displayData.getDownstreamFailureCauses().isEmpty() && displayData.getFoundFailureCauses().isEmpty()) {
+            message.append("Unknown (")
+                    .append(Jenkins.getInstance().getRootUrl())
+                    .append(displayData.getLinks().getBuildUrl())
+                    .append(")");
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/GerritMessageProviderExtension.java
```diff
@@ -98,5 +98,12 @@ private void addFailureCausesFromData(StringBuilder message, FailureCauseDisplay
             .append(displayData.getLinks().getBuildUrl())
             .append(" )");
         }
+
+        if (displayData.getDownstreamFailureCauses().isEmpty() && displayData.getFoundFailureCauses().isEmpty()) {
+            message.append("Unknown (")
```
rsandell: Why not the configured default message?

Jimilian: It's too long. UX was pretty bad: small window, a lot of text...

If you want I can create new constant for that or use default message ("No problems were identified. If you know why this problem occurred, please add a suitable Cause for it." aka `DEFAULT_NO_CAUSES_MESSAGE`).

rsandell: Well I'm mostly concerned that unfamiliar users won't know/understand what that "Unknown" comes from, but if it is the same text that appears on the build page it would be easier to understand.

TWestling: If we are going to use the standard value, then use the configured noCausesMessage, not the default one.
I could see a point in shortening it though, since "please add a suitable cause for it" isn't really helpful from Gerrit, you just want to know what went wrong. 
Maybe just "No problems were identified", then it at least looks like the default message that people are used to.

Jimilian: I decided to use "configured noCausesMessage". At least in this case users have option to change it.

Jimilian:
@TWestling, @rsandell, please, take a look :)

PR #64, opened 2017-02-21
Added a FailureCauseProvider for the MQ Notifier plugin

TWestling:
The MQ Notifier plugin sends messages to a Message Queue,
implementing the AMQP protocol, e.g. RabbitMQ, whenever a
build is finished.
The plugin allows for other plugins to add
information to the message. This commit adds support for
the Build Failure Analyzer to supply Failure Cause information
from the build.

The dependency to the MQ Notifier plugin is made optional.

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/providers/FailureCauseProvider.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/providers/FailureCauseProvider.java
@@ -53,7 +53,7 @@ public class FailureCauseProvider extends MQDataProvider {
                 failureCauseJSON.put("description", foundFailureCause.getDescription());
                 failureCauseJSON.put("categories", foundFailureCause.getCategories());
                 JSONObject foundIndicationsJSON = new JSONObject();
-                for(FoundIndication ind : foundFailureCause.getIndications()) {
+                for (FoundIndication ind : foundFailureCause.getIndications()) {
                     foundIndicationsJSON.put("pattern", ind.getPattern());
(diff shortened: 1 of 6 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/providers/FailureCauseProvider.java
```diff
@@ -0,0 +1,67 @@
...
+    public void provideCompletedRunData(Run run, JSONObject json) {
+        FailureCauseBuildAction action = run.getAction(FailureCauseBuildAction.class);
+        if (action != null) {
+            List<FoundFailureCause> foundFailureCauses = action.getFoundFailureCauses();
+            JSONObject failureCausesJSON = new JSONObject();
+            for (FoundFailureCause foundFailureCause : foundFailureCauses) {
+                JSONObject failureCauseJSON = new JSONObject();
+                failureCauseJSON.put("name",foundFailureCause.getName());
+                failureCauseJSON.put("description",foundFailureCause.getDescription());
+                failureCauseJSON.put("categories",foundFailureCause.getCategories());
```
rsandell: I don't think you can add a List of Strings like this. I think you first need to make a JSONArray don't you?

TWestling: I'll add some kind of Unit test, probably something quicker than a docker-fixture though.

Regarding adding a list of Strings like above, it works fine without first creating a JSONArray.

TWestling:
Added a commit fixing the few checkstyle warnings we had, including the ones I recently added :)

PR #65, opened 2017-02-25
Fix collecting downstream builds

Jimilian:
Current implementation of method compareTo of Run is not alighed with method `equals`.
As result `compareTo` for builds from different parents, but with same number, returns 0.
And TreeSet as any sorted set keeps only one build instead of two.

https://github.com/jenkinsci/jenkins/pull/2762

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction.java
@@ -38,6 +38,7 @@ import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
 import java.io.IOException;
+import java.util.Comparator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
@@ -342,7 +343,17 @@ public class FailureCauseBuildAction implements BuildBadgeAction {
     private static Set<Run<?, ?>> getDownstreamBuilds(
             final Run build) {
 
-        Set<Run<?, ?>> foundDbf = new TreeSet<Run<?, ?>>();
+        Set<Run<?, ?>> foundDbf = new TreeSet<Run<?, ?>>(new Comparator<Run<?, ?>>() {
+            @Override
+            public int compare(Run<?, ?> o1, Run<?, ?> o2) {
+                final int res = o1.getParent().getFullName().compareTo(o2.getParent().getFullName());
+                if (res == 0) {
+                    return o1.number - o2.number;
+                }
+
+                return res;
+            }
+        });
 
         for (DownstreamBuildFinder dbf : DownstreamBuildFinder.getAll()) {
 
--- a/src/test/java/com/sonyericsson/jenkins/plugins/bfa/DisplayDownstreamTest.java
+++ b/src/test/java/com/sonyericsson/jenkins/plugins/bfa/DisplayDownstreamTest.java
@@ -24,6 +24,7 @@ package com.sonyericsson.jenkins.plugins.bfa;
  */
 
 import com.sonyericsson.jenkins.plugins.bfa.model.FailureCause;
+import com.sonyericsson.jenkins.plugins.bfa.model.FailureCauseBuildAction;
 import com.sonyericsson.jenkins.plugins.bfa.model.FailureCauseDisplayData;
 import com.sonyericsson.jenkins.plugins.bfa.model.FailureCauseMatrixBuildAction;
 import com.sonyericsson.jenkins.plugins.bfa.model.FoundFailureCause;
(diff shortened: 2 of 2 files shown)


Jimilian:
@rsandell, please, review.

src/test/java/com/sonyericsson/jenkins/plugins/bfa/DisplayDownstreamTest.java
```diff
@@ -204,11 +260,11 @@ private void createMatrixProjectNews() throws Exception {
                 PROJECT_NE, PROJECT_EW, PROJECT_SE);
         createFreestyleProjectWithShell(PROJECT_SW, "rapakalja");
 
-        hudson.setNumExecutors(50);
+        jenkins.getInstance().setNumExecutors(50);
```
rsandell: should be `jenkins.jenkins`

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction.java
```diff
@@ -342,7 +343,17 @@ private static void adjustProjectDisplayName(
     private static Set<Run<?, ?>> getDownstreamBuilds(
             final Run build) {
 
-        Set<Run<?, ?>> foundDbf = new TreeSet<Run<?, ?>>();
+        Set<Run<?, ?>> foundDbf = new TreeSet<Run<?, ?>>(new Comparator<Run<?, ?>>() {
+            @Override
+            public int compare(Run<?, ?> o1, Run<?, ?> o2) {
+                final int res = o1.getParent().getFullName().compareTo(o2.getParent().getFullName());
```
rsandell: `Run` has a `getExternalizableId()` method you could use instead.

rsandell:
Is there a ticket for this issue?

src/test/java/com/sonyericsson/jenkins/plugins/bfa/DisplayDownstreamTest.java
Jimilian: It's not big deal, but...
```
public Jenkins getInstance() {
        return jenkins;
    }
```
How it can be affected?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction.java
Jimilian: Thanks, but in case of `getExternalizableId()` it will compare numbers in alphabetical order. I.e.
```
projectA#10
projectA#11
projectA#9
```

Jimilian:
@rsandell I'm not sure I can only suspect this one: https://issues.jenkins-ci.org/browse/JENKINS-22026
I did this fix mainly because if internal needs of my company.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction.java
rsandell: Yes but does that matter? Checking for `null` would be simpler as you would only need to check if `o1` and `o2` are `null` before proceeding.

src/test/java/com/sonyericsson/jenkins/plugins/bfa/DisplayDownstreamTest.java
rsandell: Ah, I've never used `getInstance()` myself before so I just assumed it did something else :confused: So never mind, move along :)

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction.java
Jimilian: > Yes but does that matter?

If doesn't matter for you - it doesn't matter for me as well. I assumed that initial idea of having this `TreeSet` was to have sorted list of jobs with permanent order in UI. If it's not a case, I can replace it by `HashSet` - it doesn't care about `compareTo` at all and it will fix this bug as well.

> Checking for null would be simpler as you would only need to check if o1 and o2 are null before proceeding.

But I don't need to check it in comparator.
There is already this check in place (just before pushing to `TreeSet`). Line 364.

I need this check only it's possible that `run` is not `null`, but `parent` of `run` is `null`. I didn't except that it's possible.

Jimilian: @rsandell, so, let's replace by `HashMap` and relax? :)

rsandell: Well, perhaps there is a need for having it in correct order. What matters to me is making sure there are no risk for `NullPointerException`s. You for example don't even test if `o1` or `o2` are `null`.

And then there is `getParent()` where the risk is very low that it would return `null` and `getFullName()` which is also low risk. But if any of them are it will be very difficult to fix from a stack trace since you can't tell which one is null of all those three calls on the same line.

Jimilian: @rsandell I see you point, but according Jenkins Core source code:
```java
/**
 * The project this build is for.
 */ 
public @Nonnull JobT getParent() {
    return project;
}
```
So, `@Nonnull` says that `getParent()` can not return null. Should I add check in this case? And mark `getParent()` as `@Nullable` in Jenkins Core?

> What matters to me is making sure there are no risk for NullPointerExceptions

There is a check https://github.com/jenkinsci/build-failure-analyzer-plugin/pull/65/commits/a5f5fbea7445ccec24a0c52b24b97ce9e9fe26df#diff-d0bff283a6b9423a6895c868f62a3492R363. So, if we don't have `null`s in TreeSet, if we can't put `null`, because we have check for it, and if `getParent` never returns `null` we are safe. I can add only one small thing - return `Collections.unmodifiableSet(foundDbf)` instead of `foundDbf`, if you want. But right now there is no other attempts to modify this set in BFA source code.

Jimilian: @rsandell @TWestling Guys, if you want it, I can add null-checks and suppress FindBug (or remove `@Nonnul` from core). But, please, give me some instructions - what to do to merge it :)

oleg-nenashev: @rsandell I see no risk of NPE here. If you see the code where `parent` may be really null, please point me to it

rsandell: ok, I'm probably just paranoid then :)

PR #66, opened 2017-03-06
Corrected JSON structure of RabbitMQ message

TWestling:
The found indications was only a single indication,
it is now a list.
Also, the ID of the failure cause was the key of each
failure cause, it has been moved inside the failure cause
structure.

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/providers/FailureCauseProvider.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/providers/FailureCauseProvider.java
@@ -28,6 +28,7 @@ import com.sonyericsson.jenkins.plugins.bfa.model.FoundFailureCause;
 import com.sonyericsson.jenkins.plugins.bfa.model.indication.FoundIndication;
 import hudson.Extension;
 import hudson.model.Run;
+import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import com.sonymobile.jenkins.plugins.mq.mqnotifier.providers.MQDataProvider;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/providers/FailureCauseProvider.java
```diff
@@ -46,19 +47,22 @@ public void provideCompletedRunData(Run run, JSONObject json) {
...
+                failureCauseJSON.put("id", foundFailureCause.getId());
                 failureCauseJSON.put("name", foundFailureCause.getName());
                 failureCauseJSON.put("description", foundFailureCause.getDescription());
                 failureCauseJSON.put("categories", foundFailureCause.getCategories());
-                JSONObject foundIndicationsJSON = new JSONObject();
+                JSONArray foundIndicationsJSON = new JSONArray();
                 for (FoundIndication ind : foundFailureCause.getIndications()) {
-                    foundIndicationsJSON.put("pattern", ind.getPattern());
-                    foundIndicationsJSON.put("matchingString", ind.getMatchingString());
+                    JSONObject foundIndicationJSON = new JSONObject();
```
rsandell: _nit_ With just the difference of an `s` in the variable name it makes it a bit hard to read.

rsandell: And the use of `JSON` in the name feels very redundant.

TWestling: I can agree on the naming, I'll add a new commit.

TWestling: @rsandell I feel like having JSON in the name makes it a little more easy to read since I am dealing with actual foundFailureCauses as well as json objects containing information from failurecauses/indications. I'll keep the "JSON" but make it more clear which ones are arrays and which ones are not.

PR #70, opened 2017-05-17
Specify the unit for the log size maximum (MB according to the code)

dcoraboeuf:
The maximum log size is very useful must the unit being used was not clear. I've added the unit in the description of the field.


dcoraboeuf:
Done :) It should indeed be better

PR #72, opened 2017-08-02
JENKINS-39284_Set_build_description_to_failure_cause

TreverW:
[JENKINS-39284](https://issues.jenkins-ci.org/browse/JENKINS-39284)
Add an option to the BFA configuration that, when enabled, sets the build description to the failure cause.

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
@@ -175,6 +175,19 @@ public class BuildFailureScanner extends RunListener<Run> {
                 printDownstream(buildLog, downstreamFailureCauses);
             }
 
+            if (PluginImpl.getInstance().isEnableBuildDescription() && !foundCauseList.isEmpty()) {
+              // create a build description text from the list of failures.
+              String buildDescription = "<mark><b>";
+              if (foundCauseList.get(0).getCategories() != null) {
+                buildDescription = buildDescription.concat(foundCauseList.get(0).getCategories().get(0));
+                buildDescription = buildDescription.concat(": ");
+              }
+              buildDescription = buildDescription.concat("</b> <i>");
+              buildDescription = buildDescription.concat(foundCauseList.get(0).getDescription());
+              buildDescription = buildDescription.concat("</i></mark>");
+              build.setDescription(buildDescription);
+            }
+
             StatisticsLogger.getInstance().log(build, foundCauseListToLog);
         } catch (Exception e) {
             logger.log(Level.SEVERE, "Could not scan build " + build, e);
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
@@ -133,6 +133,7 @@ public class PluginImpl extends Plugin {
 
     private int nrOfScanThreads;
     private int maxLogSize;
+    private Boolean enableBuildDescription;
 
     private Boolean graphsEnabled;
 
@@ -485,6 +486,19 @@ public class PluginImpl extends Plugin {
 
 
     /**
+     * If enableBuildDescription is enabled or not. Build Descriptions will be set to a concatenated
+     * list of the failure descriptions as a convenience.
+     * @return True if enabled.
+     */
+    public boolean isEnableBuildDescription() {
+      if (enableBuildDescription == null) {
+        return false;
+      } else {
+        return enableBuildDescription;
+      }
+    }
+
+    /**
      * Checks if the build with certain result should be analyzed or not.
      *
      * @param result the result
@@ -584,6 +598,7 @@ public class PluginImpl extends Plugin {
         testResultParsingEnabled = o.getBoolean("testResultParsingEnabled");
         testResultCategories = o.getString("testResultCategories");
         maxLogSize = o.optInt("maxLogSize");
+        enableBuildDescription = o.getBoolean("enableBuildDescription");
         int scanThreads = o.getInt("nrOfScanThreads");
         int minSodWorkerThreads = o.getInt("minimumNumberOfWorkerThreads");
         int maxSodWorkerThreads = o.getInt("maximumNumberOfWorkerThreads");
--- a/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
+++ b/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
@@ -121,5 +121,10 @@
                        value="${it.maxLogSize}"
                        default="${it.DEFAULT_MAX_LOG_SIZE}"/>
         </f:entry>
+        <f:entry title="${%Set job description to failure description}"
+             description="${%Enabling this option will set the build's description to the failure descriptions as a convenience.}">
+            <f:checkbox name="enableBuildDescription" checked="${it.enableBuildDescription}"
+                  default="false"/>
+        </f:entry>
     </f:advanced>
 </j:jelly>


src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -175,6 +175,19 @@ public static void scan(Run build, PrintStream buildLog) {
...
+              // create a build description text from the list of failures.
+              String buildDescription = "<mark><b>";
+              if (foundCauseList.get(0).getCategories() != null) {
+                buildDescription = buildDescription.concat(foundCauseList.get(0).getCategories().get(0));
+                buildDescription = buildDescription.concat(": ");
+              }
+              buildDescription = buildDescription.concat("</b> <i>");
+              buildDescription = buildDescription.concat(foundCauseList.get(0).getDescription());
+              buildDescription = buildDescription.concat("</i></mark>");
+              build.setDescription(buildDescription);
```
TWestling: The build description could already be set by another RunListener here. If we use build.getDescription first and concatenate our failure cause description, we will be safer when combining with other plugins. Later RunListeners could still just overwrite our description but we can't do much about that.

rsandell: There is also the build description plugin that writes the description in a post build step, as well as pipeline that could write a description whenever it wants during the build. So better to append than overwrite.

TreverW:
Thanks for the suggestions (this is my first OSS pull request so I'm learning how this works).  

I've appended to the build description instead of overwriting it.  I've also added some tests.  The tests are not as complete as I'd like because I'm not sure how to manipulate the build description before BFA scans the build.  I'd like to test functionality with a build description already set such as might happen from another plugin.  If you have some suggestion on how I might do that in the tests, I'd be happy to improve the tests.

rsandell:
You can either add a custom `TestBuilder` to a `FreestyleProject` or in a pipeline just do `currentBuild.description = "Hello World"`

You do something like the following to add the test builder:

```java
project.getBuildersList().add(new TestBuilder() {
    @Override
    boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
        build.setDescription("Hello World")
        return true;
    }
});
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
TreverW: Thanks for the suggestions.  I've appended the BFA description to existing description if one is set.

rsennewald:
Hello, is there any chance this can be re-reviewed and integrated with the project?  I'm looking for the exact same functionality to solve my use case and would love it if the plugin supported it out of the box and I didn't have to do a post build script in groovy to achieve this.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -466,4 +470,43 @@ private static FoundIndication findIndication(Indication indication, Run build,
...
+            }
+            if (foundCauseList.get(0).getCategories().size() > 0) {
+                buildDescription += ": ";
+            }
+        }
+
+        // Append all failure causes.
+        for (int i = 0; i < foundCauseList.size(); i++) {
+            buildDescription = buildDescription.concat("<i>");
+            buildDescription = buildDescription.concat(foundCauseList.get(i).getDescription());
```
rsandell: you guard against `null` above but not here?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -133,6 +133,7 @@
 
     private int nrOfScanThreads;
     private int maxLogSize;
+    private Boolean enableBuildDescription;
```
rsandell: _nit_ there is no need for a `Boolean` object here. `boolean`s in java defaults to `false` if not set.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -175,6 +175,10 @@ public static void scan(Run build, PrintStream buildLog) {
                 printDownstream(buildLog, downstreamFailureCauses);
             }
 
+            if (PluginImpl.getInstance().isEnableBuildDescription() && !foundCauseList.isEmpty()) {
```
rsandell: I find it strange why this would be a global only option.
But I guess adding a per project setting is a bit too much work since we don't have any else of those at the moment. So it can be added later as a `JobProperty` if needed.

src/test/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScannerHudsonTest.java
```diff
@@ -609,6 +610,120 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
     }
 
     /**
+     * Test enableBuildDescription = false does not append failure cause to build description
+     */
+    @Test
+    public void testTestEnableBuildDescriptionIfDisabled() throws Exception {
+      PluginImpl.getInstance().setTestResultParsingEnabled(true);
```
rsandell: what does test results have to do with this feature?

src/test/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScannerHudsonTest.java
```diff
@@ -609,6 +610,120 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
...
+
+        FailureCauseBuildAction action = build.getAction(FailureCauseBuildAction.class);
+        assertNotNull(action);
+
+        List<FoundFailureCause> causeListFromAction = action.getFoundFailureCauses();
+        assertEquals("Amount of failure causes does not match.", 2, causeListFromAction.size());
+        String testDescription = "Hello World<br>\n" +
+                "<mark><i>Here are details of the failure...</i>  <i>More details</i></mark>";
+
+        assertEquals(testDescription, build.getDescription());
```
rsandell: Why add test results if you don't have them in the description?

src/test/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScannerHudsonTest.java
```diff
@@ -609,6 +610,120 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
...
+        PluginImpl.getInstance().setTestResultParsingEnabled(true);
+        PluginImpl.getInstance().setEnableBuildDescription(true);
+        FreeStyleProject project = jenkins.createFreeStyleProject();
+
+        // Test with a preset build description
+        project.getBuildersList().add(new PrintToLogBuilder(BUILD_LOG));
+        project.getBuildersList().add(new TestBuilder() {
+            public boolean perform(AbstractBuild<?, ?> build, Launcher launcher,
+                                   BuildListener listener) throws InterruptedException, IOException {
+                build.getWorkspace().child("junit.xml").copyFrom(this.getClass().getResource("junit.xml"));
```
rsandell: seems unnecessary.

ssbarnea:
I like the idea but I would prefer to append-to-description option instead of overriding current value.

guidograzioli:
Also, this functionality should only be offered after enabling a global config checkbox. Changing the default behaviour without action is never a good idea.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -594,6 +595,28 @@ public int getMaxLogSize() {
     }
 
 
+    /**
+     * Set option to append failure causes to job's build description.
+     *
+     * @param buildDescriptionEnabled should build description option be turned on
+     */
+    public void setBuildDescriptionEnabled(Boolean buildDescriptionEnabled) {
```
rsandell:
```suggestion
    @DataBoundSetter
    public void setBuildDescriptionEnabled(Boolean buildDescriptionEnabled) {
```

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
```diff
@@ -87,5 +87,10 @@
         <f:entry title="${%Max size of log file}" description="${%maxLogSize}">
             <f:textbox field="maxLogSize" />
         </f:entry>
+        <f:entry title="${%Set job description to failure description}"
+             description="${%Enabling this option will set the build's description to the failure descriptions as a convenience.}">
+            <f:checkbox name="buildDescriptionEnabled" checked="${it.buildDescriptionEnabled}"
+                  default="false"/>
+        </f:entry>
```
rsandell: You should use correct databinding, either set the `field` attribute on the entry or the checkbox, one of them should work, and remove the `checked` attribute.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -534,4 +538,43 @@ private static FoundIndication findIndication(Indication indication, Run build,
...
+            buildDescription = buildDescription.concat(foundCauseList.get(i).getDescription());
+            buildDescription = buildDescription.concat("</i>");
+            if (i < (foundCauseList.size() - 1)) {
+                buildDescription += "  ";
+            }
+        }
+        buildDescription = buildDescription.concat("</mark>");
+
+        // Append this build description to any pre-existing build description
+        if (!(build.getDescription() == null) && !build.getDescription().isEmpty()) {
```
rsandell:
```suggestion
        if (StringUtilities.isNotEmpty(build.getDescription()) {
```

src/test/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScannerHudsonTest.java
TreverW: They allow a build that fails with some failure causes.  It allows the test to make sure the build description is not modified when the feature is disabled.

src/test/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScannerHudsonTest.java
TreverW: They allow a build that fails with some failure causes. It allows the test to make sure the build description is not modified when the feature is disabled.

TreverW:
Thank you for your time reviewing this PR.  I've tried to address the suggestions given.  Thank you.  I am still learning how to do plugin development for Jenkins so I appreciate the feedback.

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
TreverW: Done

stronk7:
Was looking to release notes and other random stuff when I arrived here.

Just in case it helps to anybody... we are using here since some good months ago this script, launched every X minutes via job... that exactly addresses this (set build description to the BFA causes found).

https://git.in.moodle.com/integration/nightlyjobs/-/blob/master/src/org/moodle/ci/maintenance/SetBuildDescriptionToIndications.groovy

We launch it via library, although I imagine that it can be embed without problem in a pipeline script. This s how we launch it:

```
@Library('ci') _

import org.moodle.ci.maintenance.SetBuildDescriptionToIndications

node("jenkins-maintenance") {
    stage("Run") {
        task = new SetBuildDescriptionToIndications(
            view: "${params.VIEW}",
        )
        if (task.run(this) != 0) {
            currentBuild.result = 'FAILURE'
        }
    }
}
```

You can see it in action @ https://ci.moodle.org/job/W.03.05%20-%20Behat%20-%20Chrome%20+%20Oracle/ (for example, because with Oracle, heh, is where we get more random/failures detected by BFA).

Hope it helps, ciao :-)

PR #73, opened 2017-09-04
Add statsd logging

JeanFred:
At @Arachnys we have been happily using BuildFailureAnalyzer (for about a year I think). We wanted to gain better insight on our failures using statistics, however the setup using MongoDB did not fit our needs.

This pull-request adds an option to send metrics to a Statsd server (where it can eg be flushed to Graphite).

We have been using our fork for several months now and are happy with it. We would love to see this upstreamed!

Let us know if there are changes to be made. As developer of this, @cedric-cordenier can chime-in too.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,6 +196,11 @@
             <version>1.2</version>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>com.timgroup</groupId>
+            <artifactId>java-statsd-client</artifactId>
+            <version>3.0.1</version>
+        </dependency>
         <!-- Test dependencies -->
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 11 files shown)


JeanFred:
Is there anything I can do to help this move forward?

JeanFred:
Thank you for the review @rsandell.

I will double check with @cedric-cordenier but I don’t believe we removed any existing features?
We created an abstract class `LoggingWork`, with the formerly-known `LoggingWork` renamed `KbLoggingWork` inheriting from it, alongside our new `StatsdLoggingWork`.

rsandell:
OK, I'll see if I can get some time to re-read the code, I only did a quick sweep

JeanFred:
Rebased this on top of current master.

JeanFred:
@rsandell Do you think this could potentially make it to the upstream? What would need to happen for this?

rsandell:
@JeanFred take a look at [the metrics plugin](https://wiki.jenkins.io/display/JENKINS/Metrics+Plugin) and see if you can make an integration with that on the file based knowledge base instead (should be made as an optional dependency somehow). You will get a lot out of the box there and also other users can choose their own backend.

rsandell:
Most likely superseded by #141

PR #74, opened 2017-10-17
Add a global switch to disable report when no causes identified

guidograzioli:
When a build fails and no build causes in the knowledge base are
matched against the console log, the build page reports with an
heading and a message that there was "No identified problem".

While most users would understand the message, we had reports
from the developers reporting that the message is very confusing,
in particular in the early stages after the plugin is deployed,
and the knowledge base does not include many frequent failures.

This patch introduces a global switch that lets the administrator
turn off the display of the build analyzer block on build pages
of jobs that failed and no indication was found.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,5 @@
+buildPlugin(
+    platforms: ['linux'],
+    findbugs: [run: true, archive: true],
+    checkstyle: [archive: true, unstableTotalAll: 0]
+  )
--- a/pom.xml
+++ b/pom.xml
@@ -229,7 +229,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
-                    <forkMode>always</forkMode>
+                    <forkCount>1</forkCount>
+                    <reuseForks>false</reuseForks>
                 </configuration>
             </plugin>
             <plugin>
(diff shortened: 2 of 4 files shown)


src/test/java/com/sonyericsson/jenkins/plugins/bfa/TransientCauseManagementHudsonTest.java
```diff
@@ -64,7 +65,7 @@
     public void testOnAProject() throws Exception {
         FreeStyleProject project = createFreeStyleProject("nils-job");
         project.getBuildersList().add(new Shell("env | sort"));
-        project = configRoundtrip(project);
+        project = (FreeStyleProject)configRoundtrip((Item)project);
```
guidograzioli: Had to do this here (and below) to workaround eclipse reporting "The method configRoundtrip is ambiguous for the type..." after freshly cloning the repository, like here https://github.com/jenkinsci/jenkins/commit/cabb5f010a9c450c4604f693644b685791f0fac5. Can get rid of it if requested.

src/test/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScannerHudsonTest.java
```diff
@@ -325,6 +325,31 @@ public void testNoMultilineIndicationFound() throws Exception {
     }
 
     /**
+     * Tests that "no problem identified" message is not shown when no failure cause has been found.
```
guidograzioli: Too many negations here, sorry I am not English mother-tongue and couldn't think of anything even slightly more readable.

ssbarnea:
@guidograzioli Ca you refresh it? Now we have CI enabled for this project so Jenkins will be able to verify it after you rebase it.

guidograzioli:
https://github.com/jenkinsci/build-failure-analyzer-plugin/pull/79

PR #76, opened 2017-11-02
Add Jenkinsfile to enable CI testing

ssbarnea:


ssbarnea:
@rsandell @TWestling can your review/merge it?

jglick:
It is fine but it actually needs to be proposed by someone with write access to the repository, so that the CI builder will trust them to run it. Otherwise you could wind up merging something which actually has failures (for example, no one ever tried running some tests on Windows), which then marks every other PR as broken and alarms contributors.

slide:
:+1: @ndeloof or @rsandell Do you guys have an issue with this? Can we get this merged?

ssbarnea:
Well, I think we are facing a chicken and the egg kind of issue: we don't merge because we don't know the result of testing but we cannot know the result without merging it first :)

My call would be that it should be ok to merge it like this, even with the risk of failing testing after. At least this would indicate the problem and allow others to raise new PRs that are fixing the reported issue.

slide:
The PR is being built right now, you can see the results at https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fbuild-failure-analyzer-plugin/detail/PR-77/1/pipeline

src/test/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseProjectActionHudsonTest.java
```diff
@@ -42,7 +43,10 @@
     @Test
     public void testShowLastFailureOnProjectPage() throws Exception {
         FreeStyleProject project = createFreeStyleProject();
-        project.getBuildersList().add(new Shell("test $BUILD_NUMBER -eq 2"));
+        final CommandInterpreter = Functions.isWindows()
+            ? new BatchFile("test %BUILD_NUMBER% -eq 2")
```
rsandell: is the test function really available on windows platforms?

Jenkinsfile
```diff
@@ -0,0 +1 @@
+buildPlugin()
```
rsandell: We also want to collect findbugs and checkstyle reports. And mark the build unstable if there are any checkstyle violations.
See [Gerrit Trigger](https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/Jenkinsfile) as an example.

ssbarnea: @rsandell TBH, I don't really know, I am coding blind at the moment ;) If I remember well it should be.

Do you need to approve the change in order for Jenkins to start building the updated PR?

My original intent was to disable these tests for Windows platform using something like annotations but I was not able to find an example doing that. My primary language is Python so doing anything in Java is much slower, any hints would be appreciated.

src/test/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseProjectActionHudsonTest.java
rsandell: The test can be skipped by
`assumeThat(not(Functions.isWindows()))` or `assumeThat(Functions.isUnix())` in the beginning of the test method.

Jenkinsfile
rsandell: In this case it needs to be merged in order to be tested.
I could also push it myself to a separate branch, since I'm a repo collaborator Jenkins should see my commit as trusted.
But in this case I think we'll just merge and see what happens :)

pom.xml
```diff
@@ -11,7 +11,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.580.1</version>
+        <version>2.60.3</version>
```
jglick: Not even buildable. See: https://github.com/jenkinsci/plugin-pom

src/test/java/com/sonyericsson/jenkins/plugins/bfa/DisplayDownstreamTest.java
```diff
@@ -89,6 +89,8 @@
      */
     @Test
     public void testFailureCauseDisplayData() throws Exception {
+        Assume.assumeThat(Functions.isWindows(), not(true));
```
jglick: `assumeFalse` BTW

jglick:
Please stop destroying history inside PR branches…the merger can always squash as they prefer.

pom.xml
ssbarnea: done.

ssbarnea:
@jglick ok, I will stop using force push. I am using force push only because that's the gerrit CR way but this does not apply to PRs.

ssbarnea:
@rsandell @jglick fixed and if from now it will require additional changes I will make them without force push.

Jenkinsfile
```diff
@@ -0,0 +1,5 @@
+buildPlugin(
+    jdkVersions: [8],
```
jglick: Delete; this is the default.

jglick:
```
org.junit.internal.AssumptionViolatedException: got: <false>, expected: is <true>
	at org.junit.Assume.assumeThat(Assume.java:95)
	at org.junit.Assume.assumeTrue(Assume.java:41)
	at org.junit.Assume.assumeFalse(Assume.java:48)
	at com.sonyericsson.jenkins.plugins.bfa.model.FailureCauseProjectActionHudsonTest.testShowLastFailureOnProjectPage(FailureCauseProjectActionHudsonTest.java:46)
```

jglick:
@ssbarnea are you looking at that test failure?

ssbarnea:
@jglick Kind of, as I really want to make this work. I spend a considerable amount of hours trying to prepare a Windows VM in order to run the tests there.

jglick:
Fine with me. There is an option to `buildPlugins` to specify platforms.

jglick:
Did so in #77 and the build passes.

PR #78, opened 2017-11-09
Include build parameters in json report

ssbarnea:
Include build parameters as part of statistics JSON.

JENKINS-23284

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/MongoDBKnowledgeBase.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/MongoDBKnowledgeBase.java
@@ -44,6 +44,7 @@ import com.sonyericsson.jenkins.plugins.bfa.utils.ObjectCountPair;
 import hudson.Extension;
 import hudson.Util;
 import hudson.model.Descriptor;
+import hudson.model.ParameterValue;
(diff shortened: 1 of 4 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/MongoDBKnowledgeBase.java
```diff
@@ -482,6 +484,16 @@ public void saveStatistics(Statistics stat) throws UnknownHostException, Authent
             cause.put("build", upstreamCause.getUpstreamBuild());
         }
         object.put("upstreamCause", cause);
+
+        DBObject parameters = null;
+        if (stat.getParameters() != null) {
+            parameters = new BasicDBObject();
+            for(ParameterValue item: stat.getParameters()) {
+              parameters.put(item.getName(), item.getValue());
```
rsandell: `item.getValue()` can return a complex object graph, is that really what you want? Wouldn't for example the [realized env vars](http://javadoc.jenkins.io/hudson/model/ParameterValue.html#buildEnvironment-hudson.model.Run-hudson.EnvVars-) from those parameters be better to have?

ssbarnea:
Vaya con dios Jenkins!

PR #79, opened 2017-11-09
Add a global switch to disable report when no causes identified

guidograzioli:
When a build fails and no build causes in the knowledge base are
matched against the console log, the build page reports with an
heading and a message that there was "No identified problem".

While most users would understand the message, we had reports
from the developers reporting that the message is very confusing,
in particular in the early stages after the plugin is deployed,
and the knowledge base does not include many frequent failures.

This patch introduces a global switch that lets the administrator
turn off the display of the build analyzer block on build pages
of jobs that failed and no indication was found.

### DIFF
--- a/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
+++ b/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
@@ -32,6 +32,10 @@
                  description="${%Whether to display graphs or not. NOTE: The selected storage type must support this!}">
             <f:checkbox name="graphsEnabled" checked="${it.graphsEnabled}" default="true"/>
         </f:entry>
+        <f:entry title="${%Enable text when no failure causes are found}"
+                 description="${%Whether to display the no problems identified message at all}">
+            <f:checkbox name="noCausesEnabled" checked="${it.noCausesEnabled}" default="true"/>
+        </f:entry>
         <f:entry title="${%Text when no failure causes are found}"
                  description="${%If no problems are identified, this text will be shown on the build page}">
             <f:textbox name="noCausesMessage" value="${it.noCausesMessage}"/>
--- a/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction/summary.groovy
+++ b/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction/summary.groovy
@@ -31,20 +31,27 @@ def l = namespace(lib.LayoutTagLib)
 
 index = 1
 
-tr {
-    td {
-        img(width: "48", height: "48", src: my.getImageUrl(), style: "margin-right:1em;")
-    }
-    td(style: "vertical-align: middle;") {
-        h2(_("Identified problems"))
-    }
-}
+def rootFailureDisplayData = my.getFailureCauseDisplayData()
+if (! (rootFailureDisplayData.getFoundFailureCauses().empty
+	   && rootFailureDisplayData.getDownstreamFailureCauses().empty
+	   && !PluginImpl.getInstance().isNoCausesEnabled())) {
 
-displayData(my.getFailureCauseDisplayData(), [], 0)
+	tr {
+	    td {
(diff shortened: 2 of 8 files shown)


guidograzioli:
Anybody interested in reviewing this or any other PR against this plugin? If I cannot have it merged I'll have to setup pipelines for our fork which is painful.

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
```diff
@@ -32,6 +32,10 @@
                  description="${%Whether to display graphs or not. NOTE: The selected storage type must support this!}">
             <f:checkbox name="graphsEnabled" checked="${it.graphsEnabled}" default="true"/>
         </f:entry>
+        <f:entry title="${%Enable text when no failure causes are found}"
+                 description="${%Whether to display the no problems identified message at all}">
+            <f:checkbox name="noCausesEnabled" checked="${it.noCausesEnabled}" default="true"/>
+        </f:entry>
         <f:entry title="${%Text when no failure causes are found}"
```
rsandell: Would be better ux if the text selection was an optional property. I.e. when you click the enable text checkbox the text to put in appears.

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseMatrixBuildAction/summary.groovy
```diff
@@ -33,31 +33,38 @@ def l = namespace(lib.LayoutTagLib)
...
 index = 1
 
-tr {
-    td {
-        img(width: "48", height: "48", src: my.getImageUrl(), style: "margin-right:1em;")
-    }
-    td(style: "vertical-align: middle;") {
-        h2(_("Identified problems"))
-    }
-}
```
rsandell: From what I can see the default text will never be displayed?

guidograzioli: It will when the conditional below is true (line 37); the block reappears on line 41 indented in, the unified diff does not make it clear, try to give a look with side-by-side view

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
guidograzioli: Good idea! but I'd like to keep the checkbox enabled by default (so the no indication text will appear), what do you think?

rsandell: Yes it should definitely be enabled by default. Especially for users upgrading (so try to test that just to be sure)

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction/summary.groovy
```diff
@@ -31,20 +31,27 @@ def l = namespace(lib.LayoutTagLib)
...
-        img(width: "48", height: "48", src: my.getImageUrl(), style: "margin-right:1em;")
-    }
-    td(style: "vertical-align: middle;") {
-        h2(_("Identified problems"))
-    }
-}
+def rootFailureDisplayData = my.getFailureCauseDisplayData()
+if (! (rootFailureDisplayData.getFoundFailureCauses().empty
+	   && rootFailureDisplayData.getDownstreamFailureCauses().empty
+	   && !PluginImpl.getInstance().isNoCausesEnabled())) {
```
rsandell: puh, I had to write down a truth table to make sure this was correct :) But it seems to be.
Have you done some test scenarios for this so someone don't mess it up in the future?

Could you simplify the logic somehow so it's easier to read?

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseMatrixBuildAction/summary.groovy
rsandell: Ah yes, see my comment in `FailureCauseBuildAction/summary.groovy` :)

PR #80, opened 2017-12-27
ensure PluginImpl will be ready when initializing the threadpool

Greybird:
usage of PluginImpl.getInstance in the static field can result in an exception if the plugin is not registered at the time of the call.
In the best case, the plugin is not started. In the worst one, it crashes the initialization thread, and Jenkins will wait for it forever (Jenkins core should be modified to handle this case gracefully).


rsandell:
Was there an issue number associated with this?

Greybird:
Hello @rsandell,

Not about this particular issue in this particular plugin.
However, [JENKINS-48725](https://issues.jenkins-ci.org/browse/JENKINS-48725) is related to the underlying issue that has been highlighted by this problem. Related PR is there:  jenkinsci/lib-task-reactor#3.

PR #81, opened 2018-01-15
[JENKINS-48932] - Update plugin to make it runnable with PCT

oleg-nenashev:
I was trying to make the plugin runnable with PCT in order to have a test suite for https://issues.jenkins-ci.org/browse/JENKINS-48932 . The codebase is pretty obsolete, and the tests hang on my machine IIUC. Let's see whether tests pass in CI.

@reviewbybees @rsandell

PR #83, opened 2018-01-17
[JENKINS-48932] - Update plugin to make it runable with PCT

rsandell:
Replaces #81 with test, findbugs and checkstyle fixes.

Bumped all the way up to Jenkins Core 2.7 since even that is fairly old by now and has more findbugs annotations to adhere to.

@reviewbybees

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
@@ -310,8 +310,9 @@ public class PluginImpl extends Plugin {
         PluginImpl plugin = jenkins.getPlugin(PluginImpl.class);
         if (plugin == null) {
             throw new AssertionError("Not here yet.");
+        } else {
+            return plugin;
         }
-        return plugin;
     }
 
     /**
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
@@ -503,7 +503,7 @@ public class PluginImpl extends Plugin {
      *
      * @param build the build
      * @return true if it should be scanned.
-     * @see {@link #shouldScan(Job)}
+     * @see #shouldScan(Job)
      */
     public static boolean shouldScan(Run build) {
         return shouldScan(build.getParent());
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
@@ -587,7 +587,7 @@ public class PluginImpl extends Plugin {
         int scanThreads = o.getInt("nrOfScanThreads");
         int minSodWorkerThreads = o.getInt("minimumNumberOfWorkerThreads");
         int maxSodWorkerThreads = o.getInt("maximumNumberOfWorkerThreads");
-        int thrkeepAliveTime = o.getInt("maximumNumberOfWorkerThreads");
+        int thrkeepAliveTime = o.getInt("threadKeepAliveTime");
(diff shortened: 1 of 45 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -659,7 +659,7 @@ public void setLastFailedBuildUrl() {
 
         @Override
         public String getDisplayName() {
-            return null;
+            return "";
```
oleg-nenashev: IIRC it can be already removed in this version... or not?

jglick: Just delete this override altogether.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/StatisticsLogger.java
```diff
@@ -128,12 +129,19 @@ public void run() {
             if (build instanceof AbstractBuild) {
                 AbstractBuild abstractBuild = (AbstractBuild)build;
                 Node node = abstractBuild.getBuiltOn();
-                nodeName = node.getNodeName();
+                if (node != null) {
+                    nodeName = node.getNodeName();
```
oleg-nenashev: Probably you just want to fail the run, but OK

src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/StatisticsLogger.java
```diff
@@ -128,12 +129,19 @@ public void run() {
...
+                }
             }
             int timeZoneOffset = TimeZone.getDefault().getRawOffset();
             String master;
 
-            String result = build.getResult().toString();
+
+            String result = "Unknown";
+            final Result buildResult = build.getResult();
+            if (buildResult != null) {
```
oleg-nenashev: Maybe result should be "Running" according to the getResult() docs

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -310,8 +310,9 @@ public static PluginImpl getInstance() {
         PluginImpl plugin = jenkins.getPlugin(PluginImpl.class);
         if (plugin == null) {
             throw new AssertionError("Not here yet.");
+        } else {
+            return plugin;
         }
-        return plugin;
```
oleg-nenashev: hmm... FindBugs should be fine with that

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
rsandell: Even on 1.625 it was `@Nonnull`

src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/StatisticsLogger.java
rsandell: This code shouldn't execute until after the run is done (for now) so a `null` here would mean unknown and something is wrong, but I don't want to skip logging the statistics just for that.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/StatisticsLogger.java
rsandell: It is set to unknown above.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/StatisticsLogger.java
oleg-nenashev: Any Build Step may actually change this expectation. Better safe than sorry

PR #85, opened 2018-04-20
Adding support for DynamoDB backend

citizenken:
# What
This PR implements a DynamoDB backend for FailureCause storage. It makes the following changes:
* Extends KnowledgeBase for Dynamo, inspired by the current Mongo implementation
* Adds annotations/a few extra functions to FailureCause, FailureCauseModification, and Indication to facilitate mapping models to Dynamo models/objects
* UI-base configuration options
* Extensive unit tests

From a user's perspective, the UI/modifying failures/how failures are displayed is unchanged.
From a DB perspective, I have ensured that the data structure is the same as the Mogo structure:

```
{  
   "_id":"5ad00f8160b2ee54a8330754",
   "name":"foo",
   "description":"bar",
   "comment":"",
   "indications":[  
      {  
         "@class":"com.sonyericsson.jenkins.plugins.bfa.model.indication.BuildLogIndication",
         "pattern":"baz"
      }
   ],
   "modifications":[  
      {  
         "user":"kpetti",
         "time":"2018-04-13T02:01:37.284Z"
      }
   ],
   "lastOccurred":"1970-01-01T00:00:00Z"
}
```

# Why
When evaluating the plugin for use, my team really liked the concept of a dedicated database for tracking failure info. However, without more support outside of our team, we could not justify the operationalizing/maintenance cost of a MongoDB cluster. We're not DBA's, so it falls out of our wheelhouse. Additionally, the plugin uses the mongo-java-driver 2.x, which only communicates with MongoDB v2.x, which has now been [EoL'd](https://www.mongodb.com/blog/post/mongodb-2-6-end-of-life).

For a no/low-maintenance NoSQL store, we want to use DynamoDB. It is easy to operationalize, and we don't have to dedicate much time/thought into management of the database. As our company is moving more to AWS services, it seems like a natural fit, and could be a good fit for other teams as well.

## Testing
I've written a testsuite for the new backend, as well as tests for any functionality I've added. The backend testsuite has a fairly high level of mocking, mostly because of how much DynamoDB's Java library handles behind the scenes.

## ToDo
This is a large PR because it adds the bulk of the functionality (CRUD operations). That said, there is still more to implement to reach Mongo parity:
- Caching layer for larger queries
- Statistic tracking
- Result sorting
- Investigate DynamoDB scan vs query and which is more appropriate

## UI screenshot
screen shot 2018-04-19 at 11 43 26 pm


src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -67,17 +84,30 @@
  *
  * @author Tomas Westling &lt;thomas.westling@sonyericsson.com&gt;
  */
+@DynamoDBTable(tableName = "failureCauses")
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class FailureCause implements Serializable, Action, Describable<FailureCause> {
     private static final Logger logger = Logger.getLogger(FailureCause.class.getName());
+    @DynamoDBHashKey(attributeName = "id")
+    @DynamoDBAutoGeneratedKey
```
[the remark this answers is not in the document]
rsandell: Because that could cause problems with future additions of statistics.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -589,28 +690,62 @@ public CauseManagement getAncestorCauseManagement() {
...
+    }
+
+    /**
+     * Makes checkstyle happy.
+     * @return hashcode of class
+     */
+    @Override
+    public int hashCode() {
+        //Making checkstyle happy.
+        return getClass().getName().hashCode();
```
rsandell: Same equals and hashCode problems here, but in this case just looking at the `id` might be enough.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -688,4 +823,59 @@ public AutoCompletionCandidates doAutoCompleteCategories(@QueryParameter String
...
+            List<Indication> indications = new ArrayList<>();
+            for (Map<String, String> i:toUnConvert) {
+                String clazz = i.get("@class");
+                try {
+                    // Create a new Indication instance based on the class that was stored when converting
+                    Object obj = Class.forName(clazz).getConstructor(String.class).newInstance(i.get("pattern"));
+                    Indication indication = Indication.class.cast(obj);
+                    indications.add(indication);
+                } catch (Exception e) {
+                    e.printStackTrace();
```
rsandell: Should be properly logged via for example `logger`.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/Indication.java
```diff
@@ -106,6 +106,22 @@ public String toString() {
...
+    }
+
+    /**
+     * Makes checkstyle happy.
+     * @return hashcode of class
+     */
+    @Override
+    public int hashCode() {
+        //Making checkstyle happy.
+        return getClass().getName().hashCode();
```
rsandell: dito.

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/db/DynamoDBKnowledgeBase/config.jelly
```diff
@@ -0,0 +1,38 @@
+<!--
+  ~ The MIT License
+  ~
+  ~ Copyright 2012 Sony Mobile Communications AB. All rights reserved.
```
rsandell: should be changed to yours

src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/DynamoDBKnowledgeBase.java
```diff
@@ -0,0 +1,534 @@
...
+            return items;
+        }
+
+        /**
+         * Checks that the credential file exists.
+         *
+         * @param value the pattern to check.
+         * @return {@link hudson.util.FormValidation#ok()} if everything is well.
+         */
+        public FormValidation doCheckCredentialsPath(@QueryParameter("value") final String value) {
```
rsandell: first off this should use proper credentials via the credentials store instead ( the old mongodb code was done before we had any real knowledge of how Jenkins credentials worked :) )

Second this is a potential security risk as it can be used by anyone with overall/read permission to query the existence of a file on the Jenkins master file system. So needs `@RequirePOST` and a `Jenkins.checkPermission(Jenkins.ADMINISTER)`. The jelly also needs to be updated so the check is done via POST instead of a GET.

rsandell:
Thanks for the contribution, sorry that I'm taking so much time doing reviews.

rajeev-ranjan7:
@citizenken Any update on this ? We were also looking towards DynamoDB support for this plugin

rajeev-ranjan7:
@citizenken Any update on this ? We were also looking towards DynamoDB support for this plugin

PR #87, opened 2018-08-15
Slack notifications for found failure causes

johan1252:
Commit adds Slack Notifications to the BFA plugin by making use of the Slack Notifier plugin.
**By default,**
-Slack notifications in BFA are turned off
-The slack channel specified in Slack Notifier will be used (if slack option enabled)
-All failure categories will be reported (if slack option enabled)

**The BFA plugin settings now include:**
-Toggle for turning on/off slack notifications
-Ability to specify which failure cause categories for which to notify.
-Ability to override the slack channel used for the BFA plugin

Unit tests were added in BuildFailureScannerHudsonTest for the areas of code added to the failure scanning.
As well as tests added to SlackMessageProviderTest for tests specific to the new class created.

With slack notification checkbox unchecked:
Screen Shot 2020-01-05 at 5 55 54 PM

With slack notification checkbox checked:
Screen Shot 2020-01-05 at 5 56 05 PM


src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
```diff
@@ -61,6 +61,22 @@
                  description="${%enableGerritTriggerDescription}">
             <f:checkbox name="gerritTriggerEnabled" checked="${it.gerritTriggerEnabled}" default="true"/>
         </f:entry>
+        <f:entry title="${%Send notifications to Slack}"
```
rsandell: It would be better if this whole part was an `optionalProperty` that only rendered when the slack plugin is installed.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/SlackMessageProvider.java
```diff
@@ -0,0 +1,210 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2012 Sony Mobile Communications AB. All rights reserved.
```
rsandell: Wrong copyright (should be yours) ;)

darxriggs:
@johan1252 Are you still working on this feature or should the pull request be closed unmerged?

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
johan1252: Wasn't sure how to do `optionalProperty` or `optionalBlock` depending on plugin existence.
But changed it so the additional settings are in an optionalBlock based on when checkbox for `Send Notifications to Slack` is set.

johan1252:
Have rebased the PR and addressed review comments. Sorry for the long delay!
Let me know if any concerns. Have been using this at my company for quite some time and has been useful to get notifications of environment (vs test related) related failures.

checkstyle-suppressions.xml
```diff
@@ -10,4 +10,5 @@
     <suppress checks=".+"
               files="InjectedTest.java"/>
     <suppress checks="MagicNumber" files="BuildLogIndication.java"/>
+    <suppress checks="HiddenField" files="PluginImpl.java"/>
```
rsandell: instead of hiding bad code you should try to fix them, If you have a legitimate reason to ignore an error it should be stated in the code with a `//CS IGNORE...` statement.

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
```diff
@@ -73,6 +73,23 @@
                 </f:optionalBlock>
             </table>
         </f:block>
+        <f:block>
```
rsandell: A block in a table in a block....

I think it should suffice to only have the ` <f:optionalBlock ..>` and remove `<f:block><table>`
But it's been a while since I played with that one, so I could be wrong.

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.jelly
```diff
@@ -73,6 +73,23 @@
...
+            <table>
+                <f:optionalBlock name="slackNotifEnabled"
+                                 title="${%Send notifications to Slack}"
+                                 checked="${descriptor.slackNotifEnabled}"
+                                 inline="true">
```
rsandell: One of the reasons I preferred an optional block was to not use `inline=true`   :) since all slack related data could be stored in a bean and only pollute the settings with one field, sorry if I was unclear on that part.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -240,12 +242,104 @@ public static void scan(Run build, PrintStream scanLog) {
...
+                        String category = categories.get(j);
+                        if (failureCategoryMatches(category, slackFailureCauseCategories)) {
+                            notifySlackOfFailure = true;
+                            break;
+                        }
+                    }
+                }
+            }
+            /* Create list for slack message with failure causes from build */
+            if (bufBuildFailCause.length() == 0) {
```
rsandell:
```suggestion
            if (bufBuildFailCause.isEmpty()) {
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -240,12 +242,104 @@ public static void scan(Run build, PrintStream scanLog) {
...
+            String buildName, String buildNum, String buildUrl, PrintStream scanLog) {
+        boolean notifySlackOfFailure = false;
+        StringBuffer bufBuildFailCause = new StringBuffer();
+
+        /* Check if one of the failure causes for the build matches those specified in plugin's slack settings. */
+        for (int i = 0; i < foundCauseList.size(); i++) {
+            if (!notifySlackOfAllFailures) {
+                List<String> categories = foundCauseList.get(i).getCategories();
+                if (categories != null) {
+                    for (int j = 0; j < categories.size(); j++) {
```
rsandell:
```suggestion
                    for (String category : categories) {
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -240,12 +242,104 @@ public static void scan(Run build, PrintStream scanLog) {
...
+                        String category = categories.get(j);
```
rsandell:
```suggestion
                        
```
can't suggest changes to blocks unfortunately

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -240,12 +242,104 @@ public static void scan(Run build, PrintStream scanLog) {
...
+     * @return boolean true if message successfully created, false otherwise
+     */
+    public static boolean createSlackMessage(List<FoundFailureCause> foundCauseList,
+            boolean notifySlackOfAllFailures, List<String> slackFailureCauseCategories,
+            String buildName, String buildNum, String buildUrl, PrintStream scanLog) {
+        boolean notifySlackOfFailure = false;
+        StringBuffer bufBuildFailCause = new StringBuffer();
+
+        /* Check if one of the failure causes for the build matches those specified in plugin's slack settings. */
+        for (int i = 0; i < foundCauseList.size(); i++) {
```
rsandell:
```suggestion
        for (FoundFailureCause foundCause : foundCauseList) {
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -240,12 +242,104 @@ public static void scan(Run build, PrintStream scanLog) {
...
+            if (!notifySlackOfAllFailures) {
+                List<String> categories = foundCauseList.get(i).getCategories();
```
rsandell:
```suggestion
                List<String> categories = foundCause.getCategories();
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -240,12 +242,104 @@ public static void scan(Run build, PrintStream scanLog) {
...
+                        if (failureCategoryMatches(category, slackFailureCauseCategories)) {
+                            notifySlackOfFailure = true;
+                            break;
+                        }
+                    }
+                }
+            }
+            /* Create list for slack message with failure causes from build */
+            if (bufBuildFailCause.length() == 0) {
+                bufBuildFailCause.append(foundCauseList.get(i).getName());
```
rsandell:
```suggestion
                bufBuildFailCause.append(foundCause.getName());
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -240,12 +242,104 @@ public static void scan(Run build, PrintStream scanLog) {
...
+            } else {
+                bufBuildFailCause.append("\n");
+                bufBuildFailCause.append(foundCauseList.get(i).getName());
```
rsandell:
```suggestion
                bufBuildFailCause.append(foundCause.getName());
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -240,12 +242,104 @@ public static void scan(Run build, PrintStream scanLog) {
...
+    }
+
+    /**
+     * Function to check if atleast one of the failure cause categories match the identified failure.
+     * @param category - Category of the cause for the build failure
+     * @param slackFailureCauseCategories - List of failure causes for which to notify slack
+     * @return Boolean true if atleast one category matches, false otherwise
+     */
+    private static boolean failureCategoryMatches(String category, List<String> slackFailureCauseCategories) {
+        for (int i = 0; i < slackFailureCauseCategories.size(); i++) {
```
rsandell:
```suggestion
        for (String slackCategory : slackFailureCauseCategories) {
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -240,12 +242,104 @@ public static void scan(Run build, PrintStream scanLog) {
...
+            if (category.trim().equalsIgnoreCase(slackFailureCauseCategories.get(i).trim())) {
```
rsandell:
```suggestion
            if (category.trim().equalsIgnoreCase(slackCategory.trim())) {
```

rsandell:
I just came back from xmas break, so my reviewing is a bit sluggish :)

I'll see if I can get some time to show how the optional block with check for installed slack plugin, maybe make a PR to your PR.

But I noticed similar needs to be done for the Gerrit setting, so maybe I'll do that tweak to both after merge.

guykisel:
in case anyone is googling for an easy way to add failures to slack notifications before this gets merged, here's how i do it, which might be a little hacky:
```
import groovy.transform.Field
import com.sonyericsson.jenkins.plugins.bfa.BuildFailureScanner
@Field buildScanned = false

// use the build failure analyzer to scan our build and return a list of causes and their descriptions
// needs to be @NonCPS because we access currentBuild.rawBuild
@NonCPS
def scanBuildForFailures(build=null, rescan=false)
{
    build = build ?: currentBuild.rawBuild
    def failureActions = build.getActions(com.sonyericsson.jenkins.plugins.bfa.model.FailureCauseBuildAction)
    echo "Scanning build for failures! rescan=${rescan}"
    if (rescan) {
        BuildFailureScanner.scan(build, System.out)
    } else if (failureActions.size() == 0) {
        BuildFailureScanner.scanIfNotScanned(build, System.out)
    }
    build.save()
    failureActions = build.getActions(com.sonyericsson.jenkins.plugins.bfa.model.FailureCauseBuildAction)
    def causes = []
    for (def i = 0; i < failureActions.size(); i++) {
        def failureAction = failureActions[i]
        if (!failureAction) {
            continue
        }
        def failureCauses = failureAction.getFoundFailureCauses()
        if (!failureCauses) {
            // if there's no cause, delete the cause action to avoid polluting the build object
            build.removeAction(failureAction)
            continue
        }
        for (def h = 0; h < failureCauses.size(); h++) {
            def cause = failureCauses[h]
            def causeName = cause.getName()
            def causeDesc = cause.getDescription()
            echo "Found cause: ${causeName}: ${causeDesc}"
            causes << "${causeName}: ${causeDesc}"

        }
    }
    // if we didnt find any causes, force a rescan
    if (!causes && !rescan && !buildScanned) {
        echo "No causes found, rescanning"
        causes += scanBuildForFailures(build, true)
    }
    causes.flatten()
    causes.unique()
    causes.sort()
    buildScanned = true
    return causes
}
```

you can then call this from your slack notification when a build fails and get a list of causes

chanukov:
@rsandell since this behavior matches gerrit can we allow this into master and have a follow up PR? I want to build more functionality on top of this code and would rather work off master than of a PR off master.

PR #89, opened 2018-11-02
JENKINS-47882 Upgrade mongo client to 3.x and jackson to 2.x

timja:
Note,
I've tried to upgrade dependencies but not APIs in this PR, there's deprecated code and new apis that should be used in a follow up PR.

This PR will make allowing the use of SSL simple (https://issues.jenkins-ci.org/browse/JENKINS-45315)

This may resolve https://issues.jenkins-ci.org/browse/JENKINS-49140 or at least make it much easier to achieve


src/test/java/com/sonyericsson/jenkins/plugins/bfa/db/EmbeddedMongoStatisticsTest.java
```diff
@@ -156,9 +157,11 @@ private void setUpTwoCauses() throws Exception {
      * @throws Exception if something goes wrong
      */
     @Test
+    @Ignore("Fails on new version of the mongo java driver - returns null instead of expected date")
```
timja: not sure what to do here, spent awhile on it and the existing code doesn't seem to work on the new driver.

timja: fixed

timja:
@rsandell  can you please review?

rsandell:
@timja by merging your other PRs this now needs to resolve some merge conflicts :)

timja:
done @rsandell

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
```diff
@@ -45,10 +44,6 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class FoundIndication {
 
-    /**
-     * The platform file encoding. We assume that Jenkins uses it when writing the logs.
-     */
-    protected static final String FILE_ENCODING = System.getProperty("file.encoding");
```
rsandell: could be in use by externally implemented extension point (because `protected`) So better to keep but `@Deprecate`.

timja: what could be done by it though?
This does nothing.

I've searched the jenkinsci organisation and the only external plugin using this class is the claim-plugin and its not using this
https://github.com/search?q=org%3Ajenkinsci+com.sonyericsson.jenkins.plugins.bfa.model.indication.FoundIndication&type=Code

timja: Change reverted anyway, it was unrelated

benevitygh:
After this change, I'm unable to connect to Mongo v3.4.18 , but rolling back to the previous version I can again.  Is there some sort of migration or re-configuration required to make this work?

timja:
shouldn't be, i'm using 4.x but the tests are using 3.6
https://github.com/jenkinsci/build-failure-analyzer-plugin/pull/89/files#diff-b49a3adcb5396a868ad161dea2330235R85

are you using authentication?

benevitygh:
Tried with and without, with the same results.

timja:
I'll try take a look in a couple of days, looking at the docs the version is supported:
http://mongodb.github.io/mongo-java-driver/3.8/upgrading/

timja:
What stacktrace do you get @wfreemanben ?

I tried it out, and had trouble with no authentication, but with this PR I'm able to connect fine to 3.4.18:
https://github.com/jenkinsci/build-failure-analyzer-plugin/pull/95

PR #92, opened 2018-11-14
Allow markup in no identified problems message

timja:


timja:
@rsandell small one for you

PR #94, opened 2018-12-13
Add support for Token Macro tm pipeline step

mikebro:
This PR adds the functionality to call Build Failure Analyzer from a pipeline by using the Token Macro Plugin's [`tm` pipeline step](https://jenkins.io/doc/pipeline/steps/token-macro/) that was added in v2.0, for example:
```groovy
tm 'BUILD_FAILURE_ANALYZER'
```

This pipeline step is using a new overload of `evaluate` so if you try to call it with the build failure analyzer token it won't run and it will return a message like:
> BUILD_FAILURE_ANALYZER is not supported in this context

The changes in this PR implements that overload:
```groovy
echo tm('BUILD_FAILURE_ANALYZER')
```
Output:
>[Pipeline] tm
>[BFA] Found failure cause(s):
>[BFA] Backend Compilation Error from category Backend
>[BFA] Done. 0s
>[Pipeline] echo
>Identified problems:
>* Backend Compilation Error: A build failure from MSBuild
>  * Indication 1:
>    <_link to failure_>

I've had issues with calling the macro using some parameters like `useHtmlFormat=true` -- It still runs, but without html formatting. I've been looking at the code and it seems it should work and it is most likely my unfamiliarity the Java, Groovy, and [Jenkins interesting escaping/quoting](https://gist.github.com/Faheetah/e11bd0315c34ed32e681616e41279ef4). For example the command below would return the same output as above where `useHtmlFormat` is false by default:
```
echo tm('''${BUILD_FAILURE_ANALYZER, useHtmlFormat=true}''');
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,7 +112,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>token-macro</artifactId>
-            <version>1.10</version>
+            <version>2.0</version>
             <optional>true</optional>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -112,7 +112,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>token-macro</artifactId>
-            <version>2.0</version>
+            <version>2.2</version>
             <optional>true</optional>
         </dependency>
         <dependency>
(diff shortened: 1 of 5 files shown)


mikebro:
Thank you for your feedback, I'll take a look and try to get it sorted out as soon as I can after the holidays.

mikebro:
I had some time to work on it now, like you said I was overriding the wrong overload so it was pretty straight forward to fix. I also noticed the `tm` step was added in 2.2 so I updated the package reference.

I also added some tests. The existing `TokenTest` were using `expandAll` but the [`tm` step uses expand](https://github.com/jenkinsci/token-macro-plugin/blob/master/src/main/java/org/jenkinsci/plugins/tokenmacro/TokenMacroStep.java#L44) so I made all of the existing `expandAll` tests also test that `expand` overload. I also added a `PipelineTokenTest` just to make sure bfa runs with the `tm` step.

mikebro:
I think I've covered everything, if there's anything else I can do for this PR let me know.

PR #95, opened 2019-01-18
Add empty check to password

timja:
When trying to connect with no authentication I was having issues with the jenkins UI created a secret with an empty string in it

After this change I can connect successfully


Endareth:
Change looks good to me, and the fix works in my Jenkins setup (had been broken by 1.21 released version)!

timja:
@rsandell mind reviewing please?

PR #99, opened 2019-06-01
update jackson from 2.9.3/2.9.7 to 2.9.9 CVE-2019-12086

based2:
https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9.9


rsandell:
The enforcer rules doesn't like this change.

pauxus:
Build Failure Analyzer Plugin uses jackson from the jackson 2 api plugin, so that dependency would need to be updated, not (only) the dependency management section.

PR #101, opened 2019-06-01
update mq-notifier (optional) from 1.2.1 to 1.2.8 fix CSRF vulnerability

based2:
https://mvnrepository.com/artifact/com.sonymobile.jenkins.plugins.mq/mq-notifier?repo=jenkins-releases

https://wiki.jenkins.io/display/JENKINS/MQ+Notifier+Plugin

 CSRF vulnerability and missing permission checks in MQ Notifier Plugin
SECURITY-972

Users with Overall/Read permission were able to access MQ Notifier Plugin’s form validation URL, having it connect to an attacker-specified MQ system with attacker-specified credentials.

Additionally, this form validation URL did not require POST requests, resulting in a CSRF vulnerability.

The form validation now performs a permission check and requires POST requests to be sent

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -118,7 +118,7 @@
         <dependency>
             <groupId>com.sonymobile.jenkins.plugins.mq</groupId>
             <artifactId>mq-notifier</artifactId>
-            <version>1.2.1</version>
+            <version>1.2.8</version>
             <optional>true</optional>
         </dependency>
         <dependency>


rsandell:
The security issue won't affect this plugin as we only specify the minimum version required. People will install newer versions of the plugin in their Jenkins instances and get the security update that way.

Also the enforcer rules doesn't like just bumping this (see the build log)
there is also a TODO when bumping this specific plugin to investigate other things.

PR #102, opened 2019-06-02
JENKINS-57812 Fix test connection

timja:
Previously it was always showing connection ok

Testing connection properly now with a ping:
https://stackoverflow.com/a/40690891

Fixed:
image

image

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/MongoDBKnowledgeBase.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/MongoDBKnowledgeBase.java
@@ -1022,6 +1022,11 @@ public class MongoDBKnowledgeBase extends KnowledgeBase {
      */
     private MongoClient getMongoConnection() {
         if (mongo == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/db/MongoDBKnowledgeBase.java
```diff
@@ -1022,19 +1022,22 @@ private void addIndicationsToDBObject(DBObject object, List<FoundIndication> ind
      */
     private MongoClient getMongoConnection() {
         if (mongo == null) {
+            MongoClientOptions mongoClientOptions = MongoClientOptions
+                    .builder()
+                    .connectTimeout(5000)
+                    .serverSelectionTimeout(5000)
```
timja: default timeout is 30 seconds, test connection isn't very nice with that. 5 seconds is a much better balance imo.

rsandell: Sounds OK. We might want the timeout configurable in the future, but that shouldn't be hard to add.

PR #103, opened 2019-06-02
JENKINS-42755 Missing pipeline logs

timja:
https://issues.jenkins-ci.org/browse/JENKINS-42755

Note this does get this on newer jenkins:
```
2019-06-02 10:46:55.548+0000 [id=59]	WARNING	o.j.p.workflow.job.WorkflowRun#getLogFile: Avoid calling getLogFile on test #2
java.lang.UnsupportedOperationException
```
It's covered in JENKINS-54840, there's a few places in the plugin that need updates.

In this case I haven't been able to find a way to append to a build file currently when it's not a log file on disk...

One option would be to not log to the build file anymore...


rsandell:
Have you checked that this doesn't mess with the tokenizer and sending emails? I think it might.

timja:
not sure... will take a look.

rsandell:
If it is, or if it's hard to verify, maybe keep `onCompleted` for all `AbstractBuild`s and only use `onFinalized` for any non `Abstractbuild`s
The tokenizer is not working for pipelines anyway since the emailing is done inside the build before the BFA has had a chance to run as it is, so we won't "break" anything that isn't already broken ;).

timja:
that sounds like the easiest solution 👍

timja:
I've made AbstractBuild continue to use completed, I don't use the functionality you mentioned and don't really know how to test it.

I've tested free style and pipeline, both are working as expected.
image

image

timja:
done

PR #104, opened 2019-06-02
JENKINS-54840 Remove most getLogFile references

timja:
Only remaining reference is here:
https://github.com/jenkinsci/build-failure-analyzer-plugin/compare/master...timja:feature/remove-get-log-file?expand=1#diff-7435429733d482ec50045d542da50ec5R109

Not sure the best way to handle that, I wonder if BFA should have its own log for appending to and not the main pipeline log


timja:
replaced by https://github.com/jenkinsci/build-failure-analyzer-plugin/pull/106

PR #105, opened 2019-06-02
JENKINS-55110 Fix deserialisation for mongo

timja:
https://issues.jenkins-ci.org/browse/JENKINS-55110

See:
https://github.com/mongojack/mongojack/issues/157#issuecomment-336479449


timja:
@rsandell / @TWestling could you take a look please? This is breaking all mongo new installations on the latest version

PR #108, opened 2019-07-22
New feature: fallback causes

pauxus:
Allows to designate categories as fallback. When scanning for
known causes, fallback causes are only stored if there are no non-fallback
causes present.

Use case:
- fallback cause: A maven goal failed
- specific cause: The code did not compile

Old behaviour: Both causes would be shown, leading to descriptions like
"always look for more specific causes first"

New behaviour: if a compile error is found, only that one is shown
(maven error is ignored as fallback). Otherwise, the information that
the failure was due to a maven error (along with the link!) is still
provided.

### DIFF
--- a/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.properties
+++ b/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.properties
@@ -3,3 +3,4 @@ nrOfScanThreadsDescription=Number of threads per build to use when scanning the 
 testResultParsingEnabledDescription=Treat failed test cases (as indicated by JUnit/xUnit/... publishers) as failure causes.
 testResultCategoriesDescription=A space-separated list of categories to use for failure causes representing failed test cases.
 maxLogSize=Log file with size that exceeds limit (in MB) would not be scanned, 0 - disables this check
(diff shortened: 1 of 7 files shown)


pauxus:
Renamed "generic" causes to "fallback" causes

pauxus:
Failing test fixed

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/PluginImpl/config.properties
```diff
@@ -3,3 +3,4 @@ nrOfScanThreadsDescription=Number of threads per build to use when scanning the
 testResultParsingEnabledDescription=Treat failed test cases (as indicated by JUnit/xUnit/... publishers) as failure causes.
 testResultCategoriesDescription=A space-separated list of categories to use for failure causes representing failed test cases.
 maxLogSize=Log file with size that exceeds limit (in MB) would not be scanned, 0 - disables this check
+fallbackCategoriesDescription=Causes containing on of these (space separated) categories will only be applied if there are no non-fallback causes found.
```
rsandell:
```suggestion
fallbackCategoriesDescription=Space separated list of category names that marks fallback causes. Fallback causes will only be applied if there are no non-fallback causes found.
```

I don't know if my version is better, but I had a really hard time parsing/reading the original one (maybe only because __one__ was missing an _e_) :)

PR #109, opened 2019-07-23
[JENKINS-47027] fix umlauts

pauxus:
This fixes the 'umlauts' problem (https://issues.jenkins-ci.org/browse/JENKINS-47027) which seem to be a result of wrong encoding of umlauts in +`encode()`.

By using POST instead of GET (and switching to default Descriptor based validation), we circumvent that problem.

- I moved the validation methods into the Descriptor (default behaviour)
- change the validation to POST
- Changed `FailureCauseHudsonTest` from legacy `HudsonTestCase` to using the newer `JenkinsRule` (I needed access to the WebResponseListener)

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
@@ -50,7 +50,7 @@ import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.Stapler;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
-import org.kohsuke.stapler.verb.POST;
+import org.kohsuke.stapler.interceptor.RequirePOST;
 import org.mongojack.Id;
 import org.mongojack.ObjectId;
 
@@ -199,6 +199,31 @@ public class FailureCause implements Serializable, Action, Describable<FailureCa
             }
         }
         return FormValidation.ok();
+    }
+
+    /**
+     * Form validation for {@link #description}. Checks for not empty and not "Description..."
+     *
+     * @param value the form value.
+     * @return {@link hudson.util.FormValidation#ok()} if everything is well.
+     * @deprecated Use {@link FailureCauseDescriptor#doCheckDescription(String)} instead
+     */
+    @Deprecated
+    public FormValidation doCheckDescription(@QueryParameter final String value) {
+        return getDescriptor().doCheckDescription(value);
+    }
+
+    /**
+     * Form validation for {@link #name}. Checks for not empty, not "New...", {@link Jenkins#checkGoodName(String)} and
+     * that it is unique based on the cache of existing causes.
+     *
+     * @param value the form value.
+     * @return {@link hudson.util.FormValidation#ok()} if everything is well.
+     * @deprecated Use {@link FailureCauseDescriptor#doCheckName(String, String)} instead
+     */
+    @Deprecated
+    public FormValidation doCheckName(@QueryParameter final String value) {
+        return getDescriptor().doCheckName(value, id);
     }
 
     /**
@@ -622,7 +647,7 @@ public class FailureCause implements Serializable, Action, Describable<FailureCa
          * @param value the form value.
          * @return {@link hudson.util.FormValidation#ok()} if everything is well.
          */
-        @POST
+        @RequirePOST
         public FormValidation doCheckDescription(@QueryParameter final String value) {
             if (Util.fixEmpty(value) == null) {
                 return FormValidation.error("You should provide a description.");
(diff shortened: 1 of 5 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -663,6 +616,61 @@ public String getDisplayName() {
             return "";
         }
 
+        /**
+         * Form validation for {@link #description}. Checks for not empty and not "Description..."
+         *
+         * @param value the form value.
+         * @return {@link hudson.util.FormValidation#ok()} if everything is well.
+         */
+        @POST
```
rsandell: I think `@RequirePOST` would be better suited here, so the caller gets a real error instead of a not found.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -663,6 +616,61 @@ public String getDisplayName() {
...
+        /**
+         * Form validation for {@link #name}. Checks for not empty, not "New...",
+         * {@link Jenkins#checkGoodName(String)} and
+         * that it is unique based on the cache of existing causes.
+         *
+         * @param value the form value.
+         * @param id The id (if changing an existing cause).
+         * @return {@link hudson.util.FormValidation#ok()} if everything is well.
+         */
+        @POST
```
rsandell: dito

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -199,55 +201,6 @@ public FormValidation validate(String newName,
         return FormValidation.ok();
     }
 
-    /**
-     * Form validation for {@link #description}. Checks for not empty and not "Description..."
-     *
-     * @param value the form value.
-     * @return {@link hudson.util.FormValidation#ok()} if everything is well.
-     */
-    public FormValidation doCheckDescription(@QueryParameter final String value) {
```
rsandell: Should keep an `@Deprecated` method signature here for binary compatibility that just calls the new method.
There must have been a reason we put it here in the beginning.

pauxus:
Done.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -663,6 +641,61 @@ public String getDisplayName() {
...
 
+        /**
+         * Form validation for {@link #description}. Checks for not empty and not "Description..."
+         *
+         * @param value the form value.
+         * @return {@link hudson.util.FormValidation#ok()} if everything is well.
+         */
+        @RequirePOST
+        public FormValidation doCheckDescription(@QueryParameter final String value) {
+            if (Util.fixEmpty(value) == null) {
```
jtnord: you should probably check `Item.Configure` here

pauxus: Why? The only check done here is whether the description is the default description.

rsandell: Not `Item.CONFIGUE` but `com.sonyericsson.jenkins.plugins.bfa.PluginImpl#VIEW_PERMISSION` or rather `com.sonyericsson.jenkins.plugins.bfa.PluginImpl#UPDATE_PERMISSION`.

Before when the check methods where on the object instance they where protected behind those permissions and not routable to if the user didn't have it. Now when they are moved to the descriptor they are routable through `jenkins.model.Jenkins#getDescriptorByName`.
The description check is not that important, but `doCheckName` has a potential of leaking information.

rsandell:
```suggestion
            Jenkins.getInstance().checkPermission(PluginImpl.UPDATE_PERMISSION);
            if (Util.fixEmpty(value) == null) {
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -663,6 +641,61 @@ public String getDisplayName() {
...
+         *
+         * @param value the form value.
+         * @param id The id (if changing an existing cause).
+         * @return {@link hudson.util.FormValidation#ok()} if everything is well.
+         */
+        @RequirePOST
+        public FormValidation doCheckName(
+                @QueryParameter final String value,
+                @QueryParameter final String id) {
+            if (Util.fixEmpty(value) == null) {
```
rsandell:
```suggestion
            Jenkins.getInstance().checkPermission(PluginImpl.UPDATE_PERMISSION);
            if (Util.fixEmpty(value) == null) {
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
pauxus: Added

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
pauxus: Added

pauxus:
Rebased on master

I added the permission check to the autocomplete categories method as well

PR #112, opened 2019-10-04
[JENKINS-59610 JENKINS-59593] Fixed configuration issues introduced in 1.23

TWestling:
Since 1.23, the saving of the BFA configuration is broken.
The KnowledgeBase is always overwritten.
In a LocalKnowledgeBase, this causes all causes to be removed.
In a MongoDBKnowledgeBase:
The getters for enabling statistics and successfullogging stats
were wrongly named, causing the databinding to be broken.

JENKINS-59610
JENKINS-59593

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
@@ -649,21 +649,33 @@ public class PluginImpl extends GlobalConfiguration {
 
     @Override
     public boolean configure(StaplerRequest req, JSONObject o) {
-        KnowledgeBase existingKb = this.knowledgeBase;
+        KnowledgeBase existingKb = knowledgeBase;
         req.bindJSON(this, o);
 
-        if (existingKb != null && !existingKb.equals(knowledgeBase)) {
+        if (knowledgeBase != null && !existingKb.equals(knowledgeBase)) {
+            try {
+                knowledgeBase.start();
+            } catch (Exception e) {
+                logger.log(Level.SEVERE, "Could not start new knowledge base, reverting ", e);
+                // since the knowledgebase is already overwritten via req.bindJSON, we need to
+                // restore the old if the new one can't be started
+                knowledgeBase = existingKb;
+                save();
+                return true;
+            }
             if (o.getBoolean("convertOldKb")) {
                 try {
-                    knowledgeBase.start();
                     knowledgeBase.convertFrom(existingKb);
                 } catch (Exception e) {
                     logger.log(Level.SEVERE, "Could not convert knowledge base ", e);
                 }
-                knowledgeBase.stop();
             }
+            existingKb.stop();
+        } else {
+            // Since we now overwrite the existing knowledgebase immediately, we need to put it
+            // back if it was equal to the old one, or if the new one is null.
+            knowledgeBase = existingKb;
         }
-
         save();
         return true;
     }
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
@@ -354,7 +354,7 @@ public class PluginImpl extends GlobalConfiguration {
         if (graphsEnabled == null || knowledgeBase == null) {
             return false;
         } else {
-            return knowledgeBase.isStatisticsEnabled() && graphsEnabled;
+            return knowledgeBase.isEnableStatistics() && graphsEnabled;
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/main/resources/com/sonyericsson/jenkins/plugins/bfa/db/MongoDBKnowledgeBase/config.jelly
```diff
@@ -39,10 +39,10 @@
         <f:password field="password" />
     </f:entry>
     <f:entry title="${%Enable statistics logging}">
-        <f:checkbox field="enableStatistics" default="true" />
+        <f:checkbox field="enableStatistics" default="true"/>
```
timja: I would revert the whitespace changes in this file, they're distracting too the reviewer...

rsandell:
requesting review from @timja since it probably happened while fixing the JCasC compatibility. And since it's been broken since then maybe it won't matter that we break the JCasC format?

rsandell:
@TWestling since this indicates a lack of test coverage could you add some tests for the scenarios that broke?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -649,21 +649,33 @@ public KnowledgeBase getKnowledgeBase() {
...
                 } catch (Exception e) {
                     logger.log(Level.SEVERE, "Could not convert knowledge base ", e);
                 }
-                knowledgeBase.stop();
             }
+            existingKb.stop();
+        } else {
+            // Since we now overwrite the existing knowledgebase immediately, we need to put it
+            // back if it was equal to the old one, or if the new one is null.
+            knowledgeBase = existingKb;
```
timja: not sure if this case is valid?

if it's the same one there's no need to reset it, and if it's null isn't that because the user chose to clear it?

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/db/MongoDBKnowledgeBase/config.jelly
rsandell: Turn on "Hide whitespace changes" ;)

timja:
> @TWestling since this indicates a lack of test coverage could you add some tests for the scenarios that broke?

no test coverage for the statistics bit here:
https://github.com/jenkinsci/build-failure-analyzer-plugin/blob/master/src/test/resources/com/sonyericsson/jenkins/plugins/bfa/jcasc/jcasc-local.yml / https://github.com/jenkinsci/build-failure-analyzer-plugin/blob/master/src/test/resources/com/sonyericsson/jenkins/plugins/bfa/jcasc/jcasc-mongo.yml

Was wondering why there was no change needed to the jcasc tests in this plugin

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
TWestling: My reason is that if you only save the configuration without changes to a Mongo KB, the stop and start for the cache wouldn't be run. For starting the cache, that wouldn't be a problem since it would be started as soon as you interact with the KB, but for stopping it, the updaterThread etc would be kept running right?
By putting back the saved instance, we won't have a started cache floating about.

As far as it being null, I would assume something went wrong in bindJSON. When is the actual knowledgebase set to null correctly?

timja: no idea really, I was thinking if someone cleared all the settings it might be null but not sure if that's how it behaves

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/db/MongoDBKnowledgeBase/config.jelly
TWestling: Instead of adding another commit where I change the whitespace back, I'll just leave it as it is,
it's not like the whitespace in this file makes sense anyway.

TWestling:
> @TWestling since this indicates a lack of test coverage could you add some tests for the scenarios that broke?

I've added tests for the broken use cases now.

TWestling:
@rsandell @timja You guys happy, should I merge?

rsandell:
@TWestling I have activated release drafter on this repo to make changelog authoring a bit easier. 
But this would be the first release with that, so some finageling might be needed.
So I can do the release this time if you want me to?

rsandell:
@TWestling https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.adoc

There was a checkstyle error in your test that the PR builder didn't catch. I'm fixing it in another PR that bumps jackson to 2.10 for security fix.

PR #114, opened 2019-10-10
Bump jackson2 to 2.10.0 

rsandell:
To fix security warning that GitHub irritatingly shows me all the time.
The new version of jackson-api plugin requires a new Jenkins core for unknown reason.

Also fixes a checkstyle error introduced in #112

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,12 +15,13 @@
     </parent>
 
     <properties>
-        <jenkins.version>2.73.3</jenkins.version>
+        <jenkins.version>2.138.4</jenkins.version>
         <java.level>8</java.level>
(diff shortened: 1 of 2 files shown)


rsandell:
It says the fix was done in jackson-databind 2.9.10 so I'm hoping 2.10.0 contains the same fix.

The security issue is not that severe for this plugin since users can just update the plugin in their own instance and bfa will use that version. But I think it's good to not get reported on since security scanners don't understand that aspect of jenkins plugin dependencies.

rsandell:
@MRamonLeon No it's CVE-2019-16335 and CVE-2019-14540
hmm...

PR #115, opened 2019-10-19
Fix boolean options in plugin global config

thomasgl-orange:
In the plugin global configuration, several boolean options (checkboxes) are not working: when you change their value via the GUI, save, and reload the configuration page, their state is actually not changed.

A good way to test them all is the following:
 - tick all checkboxes, save, reload, see some checkbox are still unticked
 - untick all checkboxes, save, reload, see some checkbox are still ticked

The only boolean option which works is `doNotAnalyzeAbortedJob`.  The four other ones (`graphsEnabled`, `globalEnabled`, `testResultParsingEnabled`, `gerritTriggerEnabled`) are broken.

I think this bug has been introduced by recent JCASC changes, more precisely by the adoption of `StaplerRequest#bindJSON(Object,JSONObject)` to replace the many explicit `something = o.getBoolean("something")` affectations in the implementation of `PluginImpl#configure(StaplerRequest,JSONObject)`.

Anyway, in this PR are two commits:
 - first, a test case which fails. Note that it's a bit clunky, because it has to use the actual `StaplerRequest` implementation instead of a mock (the purpose of the test being to check that `configure`, thus basically `StaplerRequest#bindJSON`, works as expected on these options). 
 - then a fix for `PluginImpl`: it changes four `setSomething(Boolean)` methods into `setSomething(boolean)`. It happens to be enough to make `bindJSON` work. Maybe it's not the only way, I don't know, but it seemed simple enough. It's a slight API change, but really, I can't think of a legitimate case for calling these setters with anything but actual boolean values (why would anything call them with `null` for instance?).

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
@@ -360,10 +360,12 @@ public class PluginImpl extends GlobalConfiguration {
 
     /**
      * Sets if graphs are enabled.
+     * Default value is false.
+     *
      * @param graphsEnabled the graph flag
      */
     @DataBoundSetter
-    public void setGraphsEnabled(Boolean graphsEnabled) {
+    public void setGraphsEnabled(boolean graphsEnabled) {
         this.graphsEnabled = graphsEnabled;
     }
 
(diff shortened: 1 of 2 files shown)


rsandell:
A `configRountrip` test should also have caught this. :man_facepalming:

PR #117, opened 2019-10-20
Add a global switch to disable report when no causes identified (PR #79 resurrection)

thomasgl-orange:
Here is a follow up for PR #79 (new option to disable the Action display when there is no detected failure cause).

It starts from that PR, and then:
 - merge `master`: I've had a few things to adapt after JCasC changes in `PluginImpl`, and in `config.jelly` I've switched to using `field` attribute for the new `f:checkbox`
 - fixed some JCasC test cases (should have been part of the previous merge commit, but oh well...)
 - addressed @rsandell comment (https://github.com/jenkinsci/build-failure-analyzer-plugin/pull/79#discussion_r152242933) about making the "no failure text" option an optional property in the config view
 - fixed the `summary.groovy` views indentation (the original PR was mixing tabs and spaces)
 - addressed @rsandell comment (https://github.com/jenkinsci/build-failure-analyzer-plugin/pull/79#discussion_r154376538) about the complex logic in `summary.groovy` views

If there is a way to contribute these commits in #79, please let me know and close this one, otherwise I guess this one can replace #79.

Note that I would have one more tiny commit to add here if #115 gets merged first (to add the new `noCausesEnabled` option to the test on global config boolean options persistence).


thomasgl-orange:
I will have a look at tests failures when I can, maybe tomorrow.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -323,6 +324,29 @@ public String getNoCausesMessage() {
...
+        }
+    }
+
+    /**
+     * Sets whether the "no indications found" message should be shown in the job page when no causes are found.
+     * Default value is true.
+     *
+     * @param noCausesEnabled on or off.
+     */
+    public void setNoCausesEnabled(boolean noCausesEnabled) {
```
rsandell:
```suggestion
    @DataBoundSetter
    public void setNoCausesEnabled(boolean noCausesEnabled) {
```

thomasgl-orange: It looks like things work fine without the annotation, see test in 80c4cfd. 

Of course, I agree the setter could be annotated (better being explicit), but then there are several other setters which also lack the same annotation (`setDoNotAnalyzeAbortedJob`, `setGlobalEnabled`, `setTestResultParsingEnabled`, `setTestResultCategories`, `setGerritTriggerEnabled`, `setNrOfScanThreads`, `setMaxLogSize`).

Do you mind if I "fix" them all in one more commit to this PR?  Or do you prefer me fixing `setNoCausesEnabled` only here, and then a new PR for the other ones?

PR #121, opened 2020-01-09
Include line number of the found indication data

GLundh:
The line number is needed for allowing us to figure out which
indicator was found first.

Note: In this PR, the linenumber will only be provided for single
line indicators. While it may be feasible to add line counting to
the multiline indicators, it's a complemetely different beast. The
multiline implementation uses a sliding search window mechanism and
line counting will incur some refactoring. If ever needed, it should
be added in a future PR.

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
@@ -49,15 +49,10 @@ public class FoundIndication {
      */
     protected static final String FILE_ENCODING = System.getProperty("file.encoding");
     private String matchingFile;
-    /**
-     * @deprecated, kept for backwards compatibility.
-     */
-    @Deprecated
-    private transient Integer matchingLine;
-
     private String pattern;
     private Run build;
     private String matchingString;
+    private Integer matchingLine;
 
     /**
      * Standard constructor.
@@ -66,13 +61,16 @@ public class FoundIndication {
      * @param originalPattern the original pattern we used to match.
      * @param matchingFile    the path to the file in which we found the match.
      * @param matchingString  the String that makes up the match.
+     * @param matchingLine    the line number of the found indication
      */
     public FoundIndication(Run build, String originalPattern,
-                           String matchingFile, String matchingString) {
+                           String matchingFile, String matchingString,
+                           Integer matchingLine) {
         this.pattern = originalPattern;
         this.matchingFile = matchingFile;
         this.build = build;
         this.matchingString = matchingString;
+        this.matchingLine = matchingLine;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
```diff
@@ -81,14 +79,17 @@ public FoundIndication(Run build, String originalPattern,
...
      * @param matchingFile the path to the file in which we found the match.
      * @param matchingString the String that makes up the match.
+     * @param matchingLine the line number of the found indication
      */
     @JsonCreator
     public FoundIndication(@JsonProperty("pattern") String pattern,
             @JsonProperty("matchingFile") String matchingFile,
-            @JsonProperty("matchingString") String matchingString) {
+            @JsonProperty("matchingString") String matchingString,
+            @JsonProperty("matchingLine") Integer matchingLine) {
```
rsandell: There might be external uses of the old constructor signature (though unlikely). To be sure we usually leave the old signature around but deprecated.

GLundh:
Hi Bobby :)

> @TWestling didn't we remove the `matchingLine` because it wasn't working properly?

It was used for anchor links in the console log, which did not work for some reason[1]. We do not intend to use it for that purpose, but for classifying issues in an external system.

[1] fd6d360840afec3cae00b1ded8f7b84b9fcdc295

rsandell:
Hi Gustaf! :)

Yes, I remember that part, but I thought it was something wonky with the line number itself, that it didn't count correctly after a while or something?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/indication/FoundIndication.java
GLundh: Ok. Brought it back, but marked it @Deprecated

GLundh: I accidentally pinged the user "deprecated" a few times due to the `@`-sign :) Sorry about that, deprecated.

GLundh: Ping @TWestling

TWestling: @rsandell Yes, my commit message says just that, but I can't remember why. I think that it had to do with the visualization of found indications when following the link from the build action. In this implementation, we aren't changing any visualization, just saving the line number for later use, so I don't think it will have much impact.

TWestling:
> Hi Gustaf! :)
> 
> Yes, I remember that part, but I thought it was something wonky with the line number itself, that it didn't count correctly after a while or something?

@rsandell If that was the case I have forgotten about it. Even if you remember correctly and we start getting the wrong line number, it shouldn't affect the plugin much in general.

GLundh:
> > Hi Gustaf! :)
> > Yes, I remember that part, but I thought it was something wonky with the line number itself, that it didn't count correctly after a while or something?
> 
> @rsandell If that was the case I have forgotten about it. Even if you remember correctly and we start getting the wrong line number, it shouldn't affect the plugin much in general.

Does this mean I can go ahead and merge?

PR #129, opened 2020-06-10
Add Categories to Slack message improve UI and fix deprecations

chanukov:
Add Failure Categories to Slack message.
Add Failure Description to slack message.
Improve UI.
Fix a case where you end up getting information about two failing cases when you are only looking to get messages for one.
Remove deprecated
java.util.concurrent.Future;
in favor of
hudson.model.queue.QueueTaskFuture;

Improve slack Message unit tests
Screen Shot 2020-06-15 at 1 05 07 PM

An example of multiple categories for a Cause:
Screen Shot 2020-07-14 at 10 53 19 AM

An example of multiple Failures in one message:
Screen Shot 2020-07-14 at 10 55 46 AM

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
@@ -279,7 +279,7 @@ public class BuildFailureScanner extends RunListener<Run> {
      * @param buildNum - Build object
      * @param buildUrl - Full URL of build
      * @param scanLog - PrintStream for the build log
-     * @return String message if message successfully created, null otherwise
+     * @return String Slack message with failure name, category and description if message successfully created, null otherwise
      */
     public static String createSlackMessage(List<FoundFailureCause> foundCauseList,
             boolean notifySlackOfAllFailures, List<String> slackFailureCauseCategories,
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
@@ -279,7 +279,8 @@ public class BuildFailureScanner extends RunListener<Run> {
      * @param buildNum - Build object
      * @param buildUrl - Full URL of build
      * @param scanLog - PrintStream for the build log
-     * @return String Slack message with failure name, category and description if message successfully created, null otherwise
+     * @return String Slack message with failure name, category and description if message successfully created,
+     * null otherwise
      */
     public static String createSlackMessage(List<FoundFailureCause> foundCauseList,
             boolean notifySlackOfAllFailures, List<String> slackFailureCauseCategories,
@@ -290,11 +291,11 @@ public class BuildFailureScanner extends RunListener<Run> {
         /* Check if one of the failure causes for the build matches those specified in plugin's slack settings. */
         for (FoundFailureCause foundCause : foundCauseList) {
             if (notifySlackOfAllFailures) {
-                //Add two new lines between found categories
+                //Add two new lines between found Failure Causes
                 if (bufBuildFailCause.length() != 0) {
                     bufBuildFailCause.append("\n\n");
                 }
-                //Create list for slack message with failure Name, Cegory and Description from build
+                //Create list for slack message with failure Name, Category and Description from build
                 bufBuildFailCause.append("*Failure Name:* ");
                 bufBuildFailCause.append(foundCause.getName());
                 bufBuildFailCause.append("\n");
@@ -310,11 +311,11 @@ public class BuildFailureScanner extends RunListener<Run> {
                     for (String category : categories) {
                         if (failureCategoryMatches(category, slackFailureCauseCategories)) {
                             notifySlackOfFailure = true;
-                            //Add two new lines between found categoreis
+                            //Add two new lines between found Failure Causes
                             if (bufBuildFailCause.length() != 0) {
                                 bufBuildFailCause.append("\n\n");
                             }
-                            // Create list for slack message with failure Name, Cegory and Description from build
+                            // Create list for slack message with failure Name, Category and Description from build
                             bufBuildFailCause.append("*Failure Name:* ");
                             bufBuildFailCause.append(foundCause.getName());
                             bufBuildFailCause.append("\n");
@@ -335,9 +336,9 @@ public class BuildFailureScanner extends RunListener<Run> {
             SlackMessageProvider slack = new SlackMessageProvider();
 
             StringBuilder s = new StringBuilder("Job *\"" + buildName + "\"*");
-            s.append(" build *#" + buildNum + "* FAILED due to following failure causes: \n");
-            s.append(bufBuildFailCause.toString() + "\nSee ");
-            s.append(buildUrl + " for details.");
+            s.append(" build *#").append(buildNum).append("* FAILED due to following failure causes: \n");
+            s.append(bufBuildFailCause.toString()).append("\nSee ");
+            s.append(buildUrl).append(" for details.");
 
             slack.postToSlack(s.toString(), scanLog);
             return s.toString();
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/Statistics.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/Statistics.java
@@ -197,19 +197,20 @@ public class Statistics {
         this.failureCauseStatisticsList = failureCauseStatistics;
     }
 
-        /**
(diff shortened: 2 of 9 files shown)


chanukov:
@johan1252 @rsandell was wondering if i could get your reviews around these modifications i made to the Slack message part of the BFA plugin.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -279,49 +279,57 @@ public static void scan(Run build, PrintStream scanLog) {
      * @param buildNum - Build object
      * @param buildUrl - Full URL of build
      * @param scanLog - PrintStream for the build log
-     * @return boolean true if message successfully created, false otherwise
+     * @return String message if message successfully created, null otherwise
```
johan1252, line 282: Java doc should be updated with what the returned string represents

src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/Statistics.java
```diff
@@ -198,7 +198,7 @@ public Statistics(@JsonProperty("projectName")    String projectName,
     }
 
         /**
-         * @deprecated, kept for backwards compatibility.
+         * deprecated, kept for backwards compatibility.
```
johan1252, line 201: Why remove `@deprecated`?

chanukov: JAVADOC complained that @deprecated is not correct and suggested to switch it. I am not set on this change and can just ignore that JAVADOC error in my IDE

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -279,49 +279,70 @@ public static void scan(Run build, PrintStream scanLog) {
      * @param buildNum - Build object
      * @param buildUrl - Full URL of build
      * @param scanLog - PrintStream for the build log
-     * @return boolean true if message successfully created, false otherwise
+     * @return String Slack message with failure name, category and description if message successfully created, null otherwise
      */
-    public static boolean createSlackMessage(List<FoundFailureCause> foundCauseList,
+    public static String createSlackMessage(List<FoundFailureCause> foundCauseList,
```
rsandell, line 284: Normally changing a public signature like this is bad. But we haven't released a version with this method in yet, so it's fine. Just a reminder I might not approve such a change in the future. ;)

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -279,49 +279,70 @@ public static void scan(Run build, PrintStream scanLog) {
...
             boolean notifySlackOfAllFailures, List<String> slackFailureCauseCategories,
             String buildName, String buildNum, String buildUrl, PrintStream scanLog) {
         boolean notifySlackOfFailure = false;
         StringBuilder bufBuildFailCause = new StringBuilder();
 
         /* Check if one of the failure causes for the build matches those specified in plugin's slack settings. */
         for (FoundFailureCause foundCause : foundCauseList) {
-            if (!notifySlackOfAllFailures) {
+            if (notifySlackOfAllFailures) {
+                //Add two new lines between found categories
```
rsandell, line 293: From what I can see this adds two new lines between each cause and not each category. Did you intend each category or each cause?

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -279,49 +279,70 @@ public static void scan(Run build, PrintStream scanLog) {
...
                     for (String category : categories) {
                         if (failureCategoryMatches(category, slackFailureCauseCategories)) {
                             notifySlackOfFailure = true;
-                            break;
+                            //Add two new lines between found categoreis
+                            if (bufBuildFailCause.length() != 0) {
+                                bufBuildFailCause.append("\n\n");
+                            }
+                            // Create list for slack message with failure Name, Cegory and Description from build
+                            bufBuildFailCause.append("*Failure Name:* ");
```
rsandell, line 318: This looks weird. If a failureCause has multiple categories you will print all it's information for each matching category.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/statistics/Statistics.java
rsandell: The switching is probably because it should be at the bottom of the javadoc and will then be missing a title.
So please fix as the description of why the deprecation is will be lost otherwise.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -279,49 +279,70 @@ public static void scan(Run build, PrintStream scanLog) {
...
         }
 
         /* Notify slack if configured to report "ALL" failures or if one of the failure categories matched. */
         if (notifySlackOfAllFailures || notifySlackOfFailure) {
             SlackMessageProvider slack = new SlackMessageProvider();
 
             StringBuilder s = new StringBuilder("Job *\"" + buildName + "\"*");
-            s.append(" build *#" + buildNum + "* FAILED due to following failure causes: \n```");
-            s.append(bufBuildFailCause.toString() + "```\nSee ");
+            s.append(" build *#" + buildNum + "* FAILED due to following failure causes: \n");
```
rsandell, line 338: Using StringBuilder to optimize string concatenations and then doing contcatenations :)
```suggestion
            s.append(" build *#").append(buildNum).append("* FAILED due to following failure causes: \n");
```
You should fix the constructor also.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -279,49 +279,70 @@ public static void scan(Run build, PrintStream scanLog) {
...
+            s.append(bufBuildFailCause.toString() + "\nSee ");
```
rsandell, line 339:
```suggestion
            s.append(bufBuildFailCause.toString()).append("\nSee ");
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
```diff
@@ -279,49 +279,70 @@ public static void scan(Run build, PrintStream scanLog) {
...
                 if (categories != null) {
                     for (String category : categories) {
                         if (failureCategoryMatches(category, slackFailureCauseCategories)) {
                             notifySlackOfFailure = true;
-                            break;
+                            //Add two new lines between found categoreis
+                            if (bufBuildFailCause.length() != 0) {
+                                bufBuildFailCause.append("\n\n");
+                            }
+                            // Create list for slack message with failure Name, Cegory and Description from build
```
rsandell, line 317:
```suggestion
                            // Create list for slack message with failure Name, Category and Description from build
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
rsandell: Where is the change that handles the return value?

chanukov: The change is only in tests, since the value was never used.
By having the value as a String I could unit test against the value returned vs just true or false.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
chanukov: I intended between failure causes will update the docs

src/main/java/com/sonyericsson/jenkins/plugins/bfa/BuildFailureScanner.java
chanukov: No it will print both Categories per Failure Cause:
for example:
Job "fail-exit1-test" build #373 FAILED due to following failure causes:
Failure Name: Fastlane Build hang
Failure Categories: [Fastlane, :apple:]
Description: fastlane finished with errors

chanukov: I have added a message to display how it looks in slack in the description

chanukov:
@rsandell I addressed your comments and fixed the things you mentioned, let me know if it looks good to you now.

PR #131, opened 2020-07-09
Fix large font-sizes after Jenkins Typography changes

stuartrowe:
- use new style classes to override heading tags

### DIFF
--- a/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction/summary.groovy
+++ b/src/main/resources/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction/summary.groovy
@@ -39,7 +39,9 @@ if (PluginImpl.getInstance().isNoCausesEnabled() || !rootFailureDisplayData.isEm
             img(width: "48", height: "48", src: my.getImageUrl(), style: "margin-right:1em;")
         }
         td(style: "vertical-align: middle;") {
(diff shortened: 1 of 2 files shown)


stuartrowe:
@rsandell please review when you have a chance

src/main/resources/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction/summary.groovy
```diff
@@ -39,7 +39,9 @@ if (PluginImpl.getInstance().isNoCausesEnabled() || !rootFailureDisplayData.isEm
             img(width: "48", height: "48", src: my.getImageUrl(), style: "margin-right:1em;")
         }
         td(style: "vertical-align: middle;") {
-            h2(_("Identified problems"))
+            h2(class: "h4") {
```
rsandell, line 42: Why not just
```suggestion
            h4(_("Identified problems"))
```
?
Seems weird to me to style a `h2` as a `h4`?

stuartrowe: For backwards compatibility. Changing the tags makes the failure indication summary text very small for Jenkins versions that don't have the typography changes. The style classes are new with the typography change so they have no effect on earlier Jenkins versions.

PR #138, opened 2021-01-13
fix currentLine number

wangyong001:
fix currentLine number issue, move it outside the indications loops
skip empty line

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureReader.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureReader.java
@@ -178,8 +178,8 @@ public abstract class FailureReader {
             String line;
             while ((line = reader.readLine()) != null) {
                 if (StringUtils.isEmpty(line)) {
-            	    currentLine++;
-            	    continue;
+                    currentLine++;
+                    continue;
                 }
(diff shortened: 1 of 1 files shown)


src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureReader.java
```diff
@@ -177,6 +177,10 @@ public FoundIndication scan(Run build, PrintStream buildLog) {
             int currentLine = 1;
             String line;
             while ((line = reader.readLine()) != null) {
+                if (StringUtils.isEmpty(line)) {
+                    currentLine++;
```
Bananeweizen, line 181: currentLine++ should be moved before the isEmpty check, and the second increment after the for-loop should be deleted.
otherwise I like this change, could save some CPU time.

malashi: I think the current implementation is actually right. If you move the `currentLine++` before the `if` and remove the one after the `for`, then you should initialize `int currentLine = 0` on line 177 instead of `1`.

rsandell:
What is the _fix currentLine number issue_? Any ticket number that can explain it for me?

rsandell:
Superseded by #152

PR #139, opened 2021-01-30
Add feature to optionally include causes in the JCasC export

allenbenz:
This adds an option to read/write causes to the JCasC config.

This is in part to workaround an issue where how I deploy jenkins causes JCasC to rewrite the plugin configurations - which erases the local knowledge base.

Made it an option as a few dozen causes would make the JCasC very long.

Some concerns I have that I was unable to resolve:
* I don't know what this will do to mongodb, I don't have a way to test that meaningfully.
* The `new Date(time)` line in `FailureCauseModification.java` is parsing `Date.toString()` output.  This _should_ be safe as long as the locale of the jenkins master doesn't change.
* The multilineBuildLogIndication JCasC "pattern" is wrong on export, it's the output of pattern.ToString() instead of the user provided one.


src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -155,6 +156,7 @@
     private String fallbackCategoriesAsString;
     private transient List<String> fallbackCategories;
 
+    private Boolean exportCausesToJCasCEnabled;
```
olivergondza, line 159: This can safely be `boolean` as that would default to `false` when old config is migrated. Consuming code can be simplified.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -629,6 +631,62 @@ public void setSlackFailureCategories(String slackFailureCategories) {
...
+     * @param causes value
+     */
+    @DataBoundSetter
+    public void setCauses(List<FailureCause> causes) {
+        if (knowledgeBase != null) {
+            for (FailureCause cause:causes) {
+                try {
+                    knowledgeBase.saveCause(cause);
+                } catch (Exception e) {
+                    logger.log(Level.WARNING, null, e);
```
olivergondza, line 664: It would be good to provide a meaningful message instead of `null` here.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -629,6 +631,62 @@ public void setSlackFailureCategories(String slackFailureCategories) {
...
+    public Collection<FailureCause> getCauses() {
+        // The Causes export can be very large so we only include it if enabled.
+        if (exportCausesToJCasCEnabled == null || !exportCausesToJCasCEnabled) {
+            return null;
+        }
+        if (knowledgeBase != null) {
+            try {
+                return knowledgeBase.getCauses();
+            } catch (Exception e) {
+                return null;
```
olivergondza, line 684: I find it unfortunate there is bare `Exception` declared as a checked exception in the api you are calling, but it is still desirable to log or propagate the exception here.

olivergondza:
Looking forward for this feature to be added in the plugin.

I have added some minor review comments.

olivergondza:
@rsandell, @TWestling, folks, can you please take a look? Thanks!

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -629,6 +631,63 @@ public void setSlackFailureCategories(String slackFailureCategories) {
         this.slackFailureCategories = slackFailureCategories;
     }
 
+    /**
+     * Get if causes from the current knowledge base should be included in the JCasC export is enabled.
+     * @return exportCausesToJCasCEnabled - on or off. null == off.
+     */
+    public boolean getExportCausesToJCasCEnabled() {
```
rsandell, line 638: Java bean standard:
```suggestion
    public boolean isExportCausesToJCasCEnabled() {
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -629,6 +631,63 @@ public void setSlackFailureCategories(String slackFailureCategories) {
...
+    }
+
+    /**
+     * Get the causes, used by JCasC to include it in the configuration export.
+     *
+     * @return the causes
+     */
+    public Collection<FailureCause> getCauses() {
+        // The Causes export can be very large so we only include it if enabled.
+        if (!getExportCausesToJCasCEnabled()) {
```
rsandell, line 677:
```suggestion
        if (!isExportCausesToJCasCEnabled()) {
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
```diff
@@ -86,13 +87,26 @@
...
+     * @param name          the name of this FailureCause.
+     * @param description   the description of this FailureCause.
      * @param comment       the comment of this FailureCause.
      * @param lastOccurred  the time at which this FailureCause last occurred.
      * @param categories    the categories of this FailureCause.
      * @param indications   the list of indications
      * @param modifications the modification history of this FailureCause.
      */
-    @DataBoundConstructor
     public FailureCause(String id, String name, String description, String comment,
```
rsandell, line 110:
```suggestion
    @Deprecated
    public FailureCause(String id, String name, String description, String comment,
```

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
```diff
@@ -629,6 +631,63 @@ public void setSlackFailureCategories(String slackFailureCategories) {
...
+     * Set the causes.
+     *
+     * @param causes value
+     */
+    @DataBoundSetter
+    public void setCauses(List<FailureCause> causes) {
+        if (knowledgeBase != null) {
+            for (FailureCause cause:causes) {
+                try {
+                    knowledgeBase.saveCause(cause);
```
rsandell, line 662: I am quite certain this will create duplicates for anything else than the file based knowledge base. Unless the `id` field is set. At least the javadoc should reflect that.

rsandell: Writing a unit test for that should be relatively easy. There are a couple of integration tests using an embedded mongodb that you can take inspiration from.

rsandell (changes requested):
Sorry for the delay in reviewing. So much to do and so little time.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
allenbenz: I've added a note about duplicates to the javadoc, there isn't much I can do about missing ids short of skipping entries that don't have one, thoughts on the preferred behavior?

allenbenz:
So I'm having a bit of a hygiene problem with the port embedded mongodb runs on.
It would be preferable to allow the OS to provide one to avoid colliding with some other process.

But I'm not able to think of a good way to inject that port number into the config, it's a resource so I can't change it at runtime, and there aren't any hooks in JenkinsConfiguredWithCodeRule or ConfigurationAsCode to apply a transform after it gets loaded but before it's parsed.

Only improvement I could think of would be to swap 27017 with something that's not the mongodb default port to lower the odds of collision but that's not a real solution.

Thoughts?

olivergondza:
@allenbenz, use `ConfigurationAsCode.get().configure()` or `ConfigurationAsCode.get().configureWith()` without the annotation. You can pre-process the input the way you like.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCause.java
rsandell: The deprecation in this scenario would only be a flag to anyone coding against it that it is not the one actually used in runtime. It can still be used in tests, just annotated with something if sportbugs or checkstyle complains.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/PluginImpl.java
rsandell: It is the Java bean standard :)

yuezhuangshi:
Looking forward for this feature to be added in the plugin.

jhaxon:
We're unable to deploy this plugin across our fleet without this functionality.

rsandell:
@jhaxon As it seems like the original author has abandoned the change anyone is welcome to continue to work on it to get it over the finish line :)

allenbenz:
@carlosgv87 unfortunately I have long since stopped working at the place where I was using Jenkins. If someone else wants to pick this up and finish it they should feel welcome to do so. My changes are MIT licensed so ownership isn't a concern for the next person.

PR #140, opened 2021-04-01
Analyze all projects in RunListener.onCompleted() scope (again)

GLundh:
In PR #103 the scope was changed for workflow jobs (and others) to
carry out the BFA analysis in RunListener.onFinalized() scope due
to JENKINS-42755.

I argue that this was the wrong way to remedy the mentioned issue
due to following factors:

1. RunListener.fireCompleted() is executed before a build is
released. When we say 'released' we mean that any waiting upstream
builds can continue (or downstream builds can be triggered).

2. RunListener.fireFinalized() is executed after the build has
been released.

The Build Failure Analyzer plugin was originally written to be
executing in RunListener.onCompleted() scope and then wait for the
analysis to be completed. This means that the waiting upstream
build will wait for its downstream builds to complete (and be
analysed) before doing its own aggregation and analysis.

Changing this behaviour is problematic, since the upstream job is
now (since PR #103) already given the OK to continue executing
before the BFA analysis is executed. And we will see a really bad
racing condition where the upstream build could complete its log
analysis before the downsteam build. This is extra problematic,
if you want to emit messages (mq-notifier plugin) with this kind
of data aggregated.

PR #103 is a complete behavioural change, which essentially hinders
aggregation (and acting upon) issues found in downstream jobs.

I argue that the correct fix should have been in the WorkflowRun,
ensuring that any error messages are printed in the log before the
RunListener.fireCompleted() is executed. This is how most (all?)
other job types handle it today.

In https://github.com/jenkinsci/workflow-job-plugin/pull/184
this behaviour was finally changed to print errors before
RunListener.onCompleted() is fired, and BFA will now be able
to pick up all issues produced in the console log from
RunListener.onCompleted().

NOTE: This PR should not be merged until a new release of the
Workflow Job Plugin is released.

More details can be found in
https://github.com/jenkinsci/workflow-job-plugin/pull/184.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,13 +11,13 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.55</version>
+        <version>4.19</version>
     </parent>
 
     <properties>
         <revision>2.0.1</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.164.3</jenkins.version>
+        <jenkins.version>2.222.4</jenkins.version>
         <java.level>8</java.level>
         <checkstyle.version>3.1.1</checkstyle.version>
         <configuration-as-code.version>1.35</configuration-as-code.version>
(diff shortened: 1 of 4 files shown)


TWestling:
It would be nice if we could include a dependency to the correct version of workflow-job ( which doesn't exist yet ) if and only if you are using workflow-job in the first place. Is there such a way to use maven dependencies? Otherwise people who update BFA without updating workflow-job will get strange behavior or we would have to introduce a new dependency to pipeline plugins.
Other than that, looks good, but then again, I am a bit biased :D

GLundh:
> It would be nice if we could include a dependency to the correct version of workflow-job ( which doesn't exist yet ) if and only if you are using workflow-job in the first place. Is there such a way to use maven dependencies? Otherwise people who update BFA without updating workflow-job will get strange behavior or we would have to introduce a new dependency to pipeline plugins.
> Other than that, looks good, but then again, I am a bit biased :D

Yes. By specifying an optional dependency should have that effect. I'm waiting for the next release of workflow-job, and I'll update this PR once that has happened.

TWestling:
Looks good, just had a nit with some commented out stuff in the pom.

GLundh:
> Looks good, just had a nit with some commented out stuff in the pom.

I cannot see any comments? Are those posted?

PR #149, opened 2021-11-26
Add metric category squashing per build feature

PayBas:
"Squash failure metrics to unique categories for each build" option can be useful when a single failed build should count towards each cause category exactly once. 

This is particularly useful when combined with the "Treat failed  test cases as failure causes." If numerous tests fail, "squash failure metrics" will prevent those categories from being incremented for every single failed test, which can skew metrics significantly.

- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue


PayBas:
Alternatively, we could make this the default behavior, which would make this entire PR about 10 lines of code ;)

rsandell:
@TWestling you forgot to set the label again ;)

PR #150, opened 2021-12-01
JENKINS-67238 ensure that we can retrieve causes and categories before the updater has run.

sfryffdc:
See https://issues.jenkins.io/browse/JENKINS-67238

This ensures that even before the Mongo KnowledgeBase is started that calls to get the causes and categories will not NullPointer.

Ideally the starting of new knowledge bases set through casc should also be addressed. However this is problematic in terms of maintaining the current functionality.

With casc the setKnowledgeBase method on the plugin is called to swap the knowledge bases. However a lot of functionality is coded in the configure method which is called when a UI config update has occurred. Attempting to move the code to start new knowledge bases into the code that is called when casc is used to update causes a lot of test failures. It is difficult to maintain the reverting to the previous knowledge base when the new one fails to start whilst moving code into the location that is common to both sources of configuration update.
 
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!


rsandell:
@TWestling you forgot to set the label for release drafter.

PR #157, opened 2022-04-22
Adapt to Token Macro's new escapeHtml parameter

rsandell:
See https://github.com/jenkinsci/token-macro-plugin/pull/118

Also added help documentation for the token macro.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!


pom.xml
```diff
@@ -364,20 +365,15 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.277.x</artifactId>
-                <version>950.v396cb834de1e</version>
+                <artifactId>bom-2.289.x</artifactId>
```
rsandell, line 368: As per usual, things needs some massage to fit into the square hole...

pom.xml
```diff
@@ -340,7 +341,7 @@
         <dependency>
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-core</artifactId>
-            <version>2.2</version>
+            <!--<version>2.2</version>-->
             <scope>test</scope>
         </dependency>
```
dwnusbaum, line 346: nit: I think this dependency already comes in transitively from `jenkins-test-harness` at this version. Can it be deleted?

pom.xml
```diff
@@ -387,7 +383,12 @@
             <dependency>
                 <groupId>com.google.code.gson</groupId>
                 <artifactId>gson</artifactId>
-                <version>2.8.6</version>
+                <version>2.8.7</version>
+            </dependency>
+            <dependency>
+                <groupId>org.jenkins-ci.main</groupId>
+                <artifactId>jenkins-test-harness</artifactId>
+                <version>1736.vc72c458c5103</version>
```
dwnusbaum, line 391: nit: Could the parent pom be updated instead? Also for what it's worth you can set the property `<jenkins-test-harness.version>` without having to declare the whole dependency.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/tokens/Renderer.java
```diff
@@ -133,7 +146,11 @@ public String render(FailureCauseBuildAction action) {
         final FailureCauseDisplayData data = action.getFailureCauseDisplayData();
         if (data.getFoundFailureCauses().isEmpty() && data.getDownstreamFailureCauses().isEmpty()) {
             logger.info("there were no causes");
-            return noFailureText;
+            if (!useHtmlFormat && escapeHtml) {
```
dwnusbaum, line 149: Would it make sense to use `!useHtmlFormat || escapeHtml` instead? Or maybe get rid of `escapeHtml` completely and just use `useHtmlFormat` to control escaping? Or in other words, if I set `escapeHtml` to `true`, would it make sense to set  `useHtmlFormat` to `true`? Or would that be a pointless configuration? Same for setting `escapeHtml` to `false` and `useHtmlFormat` to `false`.

(Maybe I'm just confused about the name `useHtmlFormat`.)

src/main/java/com/sonyericsson/jenkins/plugins/bfa/tokens/Renderer.java
```diff
@@ -297,9 +318,17 @@ private void addIndicationRepresentation(final StringBuilder stringBuilder,
             stringBuilder.append(":\n");
             stringBuilder.append(indentForDepth(indentLevel));
             stringBuilder.append(LIST_BULLET_SPACE);
-            stringBuilder.append("<");
+            if (escapeHtml) {
+                stringBuilder.append("&lt;");
```
dwnusbaum, line 322: If the plugin is adding this itself, is it always safe? Or can we not guarantee that? I'm confused about why we would emit `&lt;` here vs just not emitting anything.

dwnusbaum (commented):
I'm not really familiar with the plugin, so I added a couple of questions.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/tokens/Renderer.java
rsandell: `escapeHtml` is the parameter that I can't get rid of, it's now inherited from the parent class. `useHtmlFormat` is the one that has always been there and if it is true it will escape all relevant parts and always has. So the "new" parameter is then only relevant if the user wants a non html formatted list.

src/main/java/com/sonyericsson/jenkins/plugins/bfa/tokens/Renderer.java
rsandell: I am interpreting the situation as the user wants the text version of the output, but still want it to be safe html :man_shrugging:

pom.xml
rsandell: Yea, I just forgot to remove it after all my back and forth testing what I need to do to make the require upper bounds to be gone :)

PR #168, opened 2023-02-08
Fix ScanOnDemandVariables for Java 11 ThreadPoolExecutor.

TWestling:
As shown in JENKINS-70489, an IllegalArgumentError is thrown when scanning builds on demand.
This is because the Java 11 version of ThreadPoolExecutor is more picky with input values, making sure that corePoolSize < maximumPoolSize. This commit fixes our input values to make sure the above is true.

- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue


rsandell (approved):

TWestling:
The failing test is the one where we depend on the underlying machine to take x amount of time, which should probably get an overhaul. Nothing fails locally, only on the jenkins nodes where things take a bit longer.

PR #182, opened 2024-08-29
Require Java 17 and 2.475; adapt test suite to EE 9

basil:
Subsumes #181. Allow this plugin's test suite to be exercised regularly in Plugin Compatibility Tester (PCT) in https://github.com/jenkinsci/bom by requiring Java 17 or newer and Jenkins 2.475 (released today) or newer.

This PR goes against the usual advice from https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ by requiring a weekly release. The advantage is that this will allow plugin compatibility testing (PCT). The disadvantage is that new features and bug fixes can't be released to LTS users for another few months.

The choice is up to the maintainer as to whether or not to accept this PR. If this PR is rejected, we will stop testing this plugin in PCT.

PR #183, opened 2024-10-25
Refresh plugin for October 2024

basil:
Complete the migration started in #182.

### Testing done

`mvn clean verify`

### DIFF
--- a/src/test/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseProjectActionHudsonTest.java
+++ b/src/test/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseProjectActionHudsonTest.java
@@ -24,13 +24,16 @@
 package com.sonyericsson.jenkins.plugins.bfa.model;
 
 import hudson.Functions;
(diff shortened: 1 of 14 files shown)


src/test/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseProjectActionHudsonTest.java
```diff
@@ -43,16 +46,22 @@ public class FailureCauseProjectActionHudsonTest extends HudsonTestCase {
      */
     @Test
     public void testShowLastFailureOnProjectPage() throws Exception {
-        assumeFalse(Functions.isWindows());
```
basil, line 46: Now that we are running the test suite on Windows, fix this test so that it works on either Windows or Linux.

basil:
@TWestling This cleanup should complete the migration started in #182. There is no rush for it to be merged and released, but doing so while it is fresh in my mind would reduce risk in the (unlikely) case that there are any problems with the second half of the cleanup.

PR #417, opened 2026-04-20
fix: Icon not showing without UPDATE permission

### LINKED ISSUE #416: No Icons for non Admin-Users
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
Jenkins: 2.528.3
OS: Windows Server 2022 - 10.0
Java: 17.0.11 - Eclipse Adoptium (OpenJDK 64-Bit Server VM)
---
build-failure-analyzer:2.6.1
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

Windows Server 2022
View in Firefox 146 on Windows 11

### Reproduction steps

1. Build any Job and let it fail
2. Let BFA identify any problem

### Expected Results

The "Failed-Build" Indicator is shown. In "Builds" siedbar and in the "build details" view.

This is true for admin users.

Image
Image

### Actual Results

For Non-Admin users there is an "Image-not-loaded" indicator and the "alt"-Text showing

Image

Image

### Anything else?

_No response_

### Are you interested in contributing a fix?

_No response_

novinxy:
Seems that I introduced regression in #212 
Symbols are broken for users without UPDATE level permissions for Failure Cause

~~I fixed it by removing permissions around getIconFileName and getDisplayName in FailureCause.~~

~~For matrix builds I fixed it implementing empty getIconFileName  method.~~

I fixed it by adding new method getIcon, instead of depending on built-in getIconFileName. This will allow for easier configuration of permissions and what to display, e.g. we might not display Action in task, but still want Icon for summary

Fixes #416 

### Testing done
I did following test:
- created 2 users: admin + user
- defined simple failure cause (fail with message "error")
- created 2 projects: pipeline & matrix - both fail fast with "error"
- executed both pipelines
- checked results for both users
- icons visible for all scenarios and all users

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction.java
+++ b/src/main/java/com/sonyericsson/jenkins/plugins/bfa/model/FailureCauseBuildAction.java
@@ -73,6 +73,7 @@ public class FailureCauseBuildAction implements BuildBadgeAction {
         this.foundFailureCauses = foundFailureCauses;
     }
 
+    @Override
     public String getIconFileName() {
         if (Jenkins.getInstance().hasPermission(PluginImpl.UPDATE_PERMISSION)) {
             return PluginImpl.getDefaultIcon();
@@ -127,6 +128,11 @@ public class FailureCauseBuildAction implements BuildBadgeAction {
         return PluginImpl.getFullImageUrl("16x16", PluginImpl.DEFAULT_ICON_NAME);
     }
 
+    /**
+     * Gets the icon for the badge page.
+     *
+     * @return the icon name
+     */
     public String getIcon() {
         return PluginImpl.getDefaultIcon();
     }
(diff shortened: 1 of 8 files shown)


novinxy:
Hi @rsandell could I ask you for a review ?

rsandell:
Well it used to be at least that if those methods returns null the action doesn't get listed in the UI. I'm not sure if that is relevant or not for these particular actions?

novinxy:
> Well it used to be at least that if those methods returns null the action doesn't get listed in the UI. I'm not sure if that is relevant or not for these particular actions?

I didn't knew about it! Thanks for pointing that out!
I checked and it actually creates some strange artifact for Matrix build
image


I read documentation it all getIconFileName getDisplayName getUrlName should return null
But it's kind of funny as for non-matrix FailureCause those are configured for managemnet of the cause...
At least for me it looks strange that it's both on build and pipeline level
At least for me, it would make more sense if those were only added at the job/pipeline level (I'll check if it's possible to fix it in this PR)

image
image

I'll als omake sure for normal freestyle job it looks correct

novinxy:
@rsandell 
I decided to just implement new getIcon method.
It will allow for easy configuration of what to display (I think it will come in handy as I want to fix/simplify in next PR what's displayed in Task  List, e.g. there is Failure Management which doesn't work for matrix run)
It also simplified change as I didn't need to modify any permissions etc.
It works for all builds and all permissions

Only downside is that I renamed FailureCauseColumn.getIconFileName to getIcon. I introduced that function quite recently and I didn't want to add unnecessary new implementations and depredations (as I wrote it's quite new and it's column so low chance someone will depend on it).
Should I leave this rename or deprecate it and add new function ?

rsandell (approved):

